PR #16, opened 2016-12-20
Refine README.md

arturdryomov:


README.md
```diff
@@ -1,115 +1,101 @@
...
-Put the following content in your local `local.properties` file.
+  We recommend you to subscribe to changes in this repo somehow (follow it on GitHub / watch for tweets of its maintainers / etc). This will allow you to apply best practises we found to make your Remote Build better, faster and safer.
 
-```properties
-# Required.
-remote_build.machine={BUILD_MACHINE_NAME}
-```
+2. Update your `local.properties` file.
 
-That'll be passed to `ssh` as parameter, `user@machine` or `machine` or `ip` will be ok (depending on your ssh and remote machine config of course).
```
artem-zinnatullin: I believe we should keep this example to avoid confusion

README.md
```diff
@@ -147,56 +132,56 @@ $ ./gradlew assembleDebug
...
 
-## Use best hardware available for remote machine
+## Use The Best Hardware Available for Remote Machine
 
-Performant CPU (more cores, higher frequency, more cache), fast SSD, fast RAM (~8 GB per user), fast network.
+Performant CPU (more cores, higher frequency, more cache), fast SSD, fast RAM (~8 GB per user), fast network will help a lot.
 
-License
-=======
+We’ve created this with a simple motivation in mind — it is always easier to beef up a single headless Linux-based PC hardware unit instead of paying a ridiculous amount of money to some companies (cough-cough) still using mobile CPUs in their notebooks. Use and abuse this concept as you wish.
```
artem-zinnatullin: Haha

artem-zinnatullin (changes requested):
Small changes pls

README.md
dhartwich1991: 😀

README.md
ming13: We introduce `BUILD_MACHINE_NAME` just before and use it afterwards as a credential (kind of) to put to SSH config. After putting it to config it should be used there.  Not sure it is necessary to repeat ourselves.

artem-zinnatullin: Let's add a sample project to the repo then!

artem-zinnatullin (approved):

PR #25, opened 2016-12-21
Add changelog for v1.0.1

artem-zinnatullin:


CHANGELOG.md
```diff
@@ -0,0 +1,13 @@
+#mainframer Change Log
```
ming13: Omit project name, it is not necessary.

CHANGELOG.md
```diff
@@ -0,0 +1,13 @@
...
+
+## Version 1.0.1
```
ming13: I would prefer [Square format](https://github.com/square/retrofit/blob/master/CHANGELOG.md) with dates in subsections.

PR #40, opened 2016-12-28
We are now able to specify a custom port for the remote connection

kokeroulis:
It would be nice if we could specify the custom port in which we want to connect using the ssh.
There are some cases that the default port is not available.

### DIFF
--- a/mainframer.sh
+++ b/mainframer.sh
@@ -15,6 +15,7 @@ PROJECT_DIR_NAME="$( basename "$PROJECT_DIR")"
 REMOTE_BUILD_MACHINE=$(awk -F "=" '/remote_build.machine/ {print $2}' "$PROJECT_DIR/local.properties")
 LOCAL_COMPRESS_LEVEL=$(awk -F "=" '/remote_build.local_gzip_level/ {print $2}' "$PROJECT_DIR/local.properties")
 REMOTE_COMPRESS_LEVEL=$(awk -F "=" '/remote_build.local_gzip_level/ {print $2}' "$PROJECT_DIR/local.properties")
+REMOTE_SSH_PORT=$(awk -F "=" '/remote_build.ssh_port/ {print $2}' "$PROJECT_DIR/local.properties")
 
 if [ -z "$LOCAL_COMPRESS_LEVEL" ]; then
 	LOCAL_COMPRESS_LEVEL=1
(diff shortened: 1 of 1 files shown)


artem-zinnatullin:
You can do this through `~/.ssh/config`

mainframer.sh
```diff
@@ -46,10 +51,10 @@ rsync --archive --delete --compress-level=$LOCAL_COMPRESS_LEVEL \
 --exclude='captures' \
 --exclude='**/build' \
 --exclude='**/local.properties' \
---rsh "ssh" ./ "$REMOTE_BUILD_MACHINE:~/$PROJECT_DIR_NAME"
+--rsh "ssh -p$REMOTE_SSH_PORT" ./ "$REMOTE_BUILD_MACHINE:~/$PROJECT_DIR_NAME"
```
artem-zinnatullin: Space between `-p` and `$REMOTE_SSH_PORT`?

mainframer.sh
```diff
@@ -62,7 +67,7 @@ rsync --archive --delete --compress-level=$REMOTE_COMPRESS_LEVEL \
 --exclude='artifacts' \
 --exclude='captures' \
 --exclude='**/local.properties' \
---rsh "ssh" "$REMOTE_BUILD_MACHINE:~/$PROJECT_DIR_NAME/" ./
+--rsh "ssh -p$REMOTE_SSH_PORT" "$REMOTE_BUILD_MACHINE:~/$PROJECT_DIR_NAME/" ./
```
artem-zinnatullin: Same as above

artem-zinnatullin (changes requested):

artem-zinnatullin:
Not sure I'd like to include this since `~/.ssh/config` already allows you specify that and we actually already recommend to create separate config for remote build machine in `~/.ssh/config` to speedify connection so it's already used by most users and doesn't looks like required feature in `mainframer.sh` itself…

@ming13 thoughts?

kokeroulis:
@artem-zinnatullin i didn't know about ~/.ssh/config, so it looks reasonable to me...

artem-zinnatullin:
@kokeroulis general rule before spending time on creating pull request — search & create an issue :) 

i_know_that_feel_bro_by_rober_raik-d4cxn5a

ming13:
Yep, we should delegate it to the transport itself. As a bonus, you will a get a configured SSH connection, i. e. there will be an ability to SSH to the machine without specifying the port (if I remember it correctly).

artem-zinnatullin:
Yup. Sorry @kokeroulis, but I'm closing this!

kokeroulis:
@artem-zinnatullin np!

PR #47, opened 2016-12-30
Better local.properties parsing

dmitry-novikov:


artem-zinnatullin:
@dmitry-novikov btw, where is your avatar?

mainframer.sh
```diff
@@ -11,10 +11,14 @@ DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 PROJECT_DIR=$DIR
 PROJECT_DIR_NAME="$( basename "$PROJECT_DIR")"
 
+function prop {
```
ming13: `property`

Issue #50, opened 2017-01-03
Extract config parameters to a separate file instead of using local.properties


artem-zinnatullin:
Since 2.x will already be breaking #19, we can also extract mainframer config parameters to a separate file since Android Studio adds not very funny header to `local.properties` (though I've never saw such behavior):

```
## This file is automatically generated by Android Studio.
# Do not modify this file -- YOUR CHANGES WILL BE ERASED!
```

Something like `.mainframerconfig` should be fine I think.

ibit:
@artem-zinnatullin the config file option should be probably updated in the README file

artem-zinnatullin:
Sure, we'll update README before v2.0.0 release, it's still in development at the moment, created #72 to track this.

PR #51, opened 2017-01-03
Extract exclude rules to separate files, abstract away from Gradle.

### LINKED ISSUE #19: Extract local and remote excludes as configurable parameters
lines like 
```--exclude='artifacts' \```
or 
```--exclude='kotlin' \``` 
show that it is still bound to a particular project. 

It might me a good idea to extract it. 

As separate files?
```.localignore```
```.remoteignore```
?

### PR DESCRIPTION
**Pull request by artem-zinnatullin**
Closes #19.

This allows `mainframer.sh` to support basically *any* build system, see `samples/gradle` for gradle example.

### DIFF
--- a/mainframer.sh
+++ b/mainframer.sh
@@ -15,6 +15,8 @@ function property {
     grep "^${1}=" $PROJECT_DIR/local.properties | cut -d'=' -f2
 }
 
+pushd "$PROJECT_DIR"
+
 # Read config variables from local.properties.
 REMOTE_BUILD_MACHINE=$(property 'remote_build.machine')
 LOCAL_COMPRESS_LEVEL=$(property 'remote_build.local_gzip_level')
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: mainframer.sh [hunk 1]
```diff
@@ -40,32 +42,39 @@ if [ -z "$BUILD_COMMAND" ]; then
 	exit 1
 fi
 
-pushd "$PROJECT_DIR"
+# Read local exclude rules.
+LOCAL_EXCLUDE_FILE=".mainframerignorelocal"
```
dmitry-novikov: What if we name this files `.mfignore.local` and `.mfignore.remote`?

artem-zinnatullin: I'm thinking about separate folder for mainframer configs where we could place .ignorelocal,  .ignoreremote and #50 .config

mainframer.sh
```diff
@@ -40,32 +61,21 @@ if [ -z "$BUILD_COMMAND" ]; then
 	exit 1
 fi
 
-pushd "$PROJECT_DIR"
+# Read exclude rules.
+LOCAL_EXCLUDE="$(readExcludesFile .mainframerignorelocal)"
```
ming13: Not sure about that. What about `.mainframerlocalignore`?

mainframer.sh
```diff
@@ -40,32 +61,21 @@ if [ -z "$BUILD_COMMAND" ]; then
...
-rsync --archive --delete --compress-level=$LOCAL_COMPRESS_LEVEL \
---exclude='.gradle' \
---exclude='.idea' \
---exclude='**/.git/' \
---exclude='artifacts' \
---exclude='captures' \
---exclude='**/build' \
---exclude='**/local.properties' \
---rsh "ssh" ./ "$REMOTE_BUILD_MACHINE:~/$PROJECT_DIR_NAME"
+eval "rsync --archive --delete --compress-level=$LOCAL_COMPRESS_LEVEL $LOCAL_EXCLUDE --rsh ssh ./ $REMOTE_BUILD_MACHINE:~/$PROJECT_DIR_NAME"
```
ming13: At this point we can make this a function with four arguments. Not sure if it’s worth it.

samples/gradle/.mainframerignoreremote
```diff
@@ -0,0 +1,5 @@
+.gradle
```
ming13: Samples don’t entirely repeat what we have at this point.

mainframer.sh
```diff
@@ -40,32 +61,21 @@ if [ -z "$BUILD_COMMAND" ]; then
...
+
 # Sync project to remote machine.
-rsync --archive --delete --compress-level=$LOCAL_COMPRESS_LEVEL \
---exclude='.gradle' \
---exclude='.idea' \
---exclude='**/.git/' \
---exclude='artifacts' \
---exclude='captures' \
---exclude='**/build' \
---exclude='**/local.properties' \
```
ming13: Remote and local excludes look very similar. `.mainframerignorecommon`?

artem-zinnatullin: No, that's totally up to each project members

mainframer.sh
```diff
@@ -15,6 +15,27 @@ function property {
     grep "^${1}=" $PROJECT_DIR/local.properties | cut -d'=' -f2
 }
 
+function readExcludesFile {
```
artem-zinnatullin: Nothing, will try

samples/gradle/.mainframerignoreremote
artem-zinnatullin: Yup, it's basic `gradle` sample, not `gradle-android`

ming13: `gradle-android` will make more sense since our userbase mostly consists of Android developers (for now).

artem-zinnatullin: Will do in a separate pr.

mainframer.sh
artem-zinnatullin: See #54. I'd vote for current names at the moment and then if we decide to move configs into a separate folder -> rename this files, names are not the point of this PR.

mainframer.sh
artem-zinnatullin: Though, ok, this looks like a valid feature, let's think about it and add later if we'll agree on it

artem-zinnatullin:
Switched to rsync builtin `--exclude-from`, PTAL!

I'd vote for merging as is and adding other things in separate PRs.

mainframer.sh
artem-zinnatullin: No need for now

PR #52, opened 2017-01-03
Update README.md

meierjan:
added note for task pitfall


meierjan:
I just wanted to point this out as this is a stupid pitfall for starters.

artem-zinnatullin:
Ah damn, right right. We don't have flavors, so it didn't come up in our use case.

artem-zinnatullin:
Merged as is, but we'll reword that in some future,  thanks!

meierjan:
Of course. Thanks!

PR #65, opened 2017-01-06
Handle non-existing ignore files.

artem-zinnatullin:
`rsync` exits with error code if file passed to `--ignore-from` does not exist.

This PR handles that and also refactors `mainframer.sh` into bunch of functions for easier understanding.


artem-zinnatullin:
PAUSEMERGE

Actually, let's merge test framework #66 first and then I'll add tests for this PR.

artem-zinnatullin:
Rebased and added tests for ignore rules!

artem-zinnatullin:
RESUMEMERGE, @ming13 PTAL

PR #66, opened 2017-01-07
Add test suite.

### LINKED ISSUE #8: Add CI
We can try to `ssh` to the same machine that runs CI build (Travis, CircleCI, etc) and then run some Gradle build and check its results.

artem-zinnatullin:
Closes #8. 

PR adds basic tests and "test runner" to the project. Design is pretty straightforward, I've added support for `ssh`ing any machine and for now it's configured to run against `localhost` as a "remote build machine", but if we'll find out that neither Travis nor CircleCI allow us run `sshd`, we'll setup separate machine and then we'll be able to encrypt ssh key for connection from Travis or CircleCI using their apis for encryption.

I don't have permissions to add CI like Travis or CircleCI ~~fluck-fluck~~, so we'll figure out that later in a separate PR.

For now we can run test suite locally (for each PR): `bash test.sh`, this is how it looks like:

screen shot 2017-01-07 at 04 28 13

cc @dmitry-novikov PTAL (not blocking reviewer).


ming13:
There is a medication for what you have, my friend.

test/test_keeps_files_on_local_machine.sh
```diff
@@ -0,0 +1,26 @@
...
+# Execute common pre-setup, include test functions.
+source "$DIR/common.sh"
+
+printTestStarted
+
+# Create several files that should be synced to remote machine AND KEEPT after execution.
+mkdir "$BUILD_DIR/src"
+touch "$BUILD_DIR/src/file1.txt"
+touch "$BUILD_DIR/src/file2.txt"
+touch "$BUILD_DIR/src/file3.txt"
```
ming13: Can we iterate over files as a list somehow?

test/test_moves_files_to_local_machine.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+set -e
+
+# You can run it from any directory.
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
```
ming13: Sounds like a repetition.

test/test_keeps_files_on_local_machine.sh
artem-zinnatullin: Sure, but I didn't want to complicate source code, let's leave as is for easier readability by us and others?

test/test_moves_files_to_local_machine.sh
artem-zinnatullin: Yup, but the problem is that I can't extract this to `common.sh` because to include `common.sh` I need to know directory where the script is in a way that does not depend on from where was the script launched…

test/common.sh
```diff
@@ -0,0 +1,82 @@
+#!/bin/bash
+set -e
+
+if [ "$DEBUG_MODE_FOR_ALL_TESTS" == "true" ]; then
```
dmitry-novikov: `VERBOSE_TESTS`?

test/test_moves_files_to_local_machine.sh
dmitry-novikov: @artem-zinnatullin, what is we `cd` to `test` folder before looping through each test and on test end `pop` original folder?

dmitry-novikov: Or we can run test like `( cd "$DIR" ; ./"$test_" )`.

artem-zinnatullin: I'd like to be able to run each test separately from any directory simply by `bash path/test.sh`

PR #75, opened 2017-01-09
Use .mainframer/personalconfig instead of local.properties.

### LINKED ISSUE #50: Extract config parameters to a separate file instead of using local.properties

artem-zinnatullin:
Closes #50.

In future we can add `.mainframer/teamconfig` that usually meant to be commited to VCS while `.mainframer/personalconfig` is meant to be ignored from VCS.

### DIFF
--- a/mainframer.sh
+++ b/mainframer.sh
@@ -11,17 +11,23 @@ DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 PROJECT_DIR=$DIR
 PROJECT_DIR_NAME="$( basename "$PROJECT_DIR")"
 MAINFRAMER_DIR="$PROJECT_DIR/.mainframer"
+PERSONAL_CONFIG_FILE="$MAINFRAMER_DIR/personalconfig"
 IGNORE_LOCAL_FILE="$MAINFRAMER_DIR/localignore"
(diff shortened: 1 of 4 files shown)


mainframer.sh
```diff
@@ -32,7 +38,7 @@ if [ -z "$REMOTE_COMPRESS_LEVEL" ]; then
 fi
 
 if [ -z "$REMOTE_BUILD_MACHINE" ]; then
-	echo "Please specify remote build machine in local.properties"
+	echo "Please specify \"$REMOTE_MACHINE_PROPERTY\" in $PERSONAL_CONFIG_FILE"
```
arturdryomov, line 41: Single quotes?

arturdryomov (approved):

mainframer.sh
artem-zinnatullin: Nah,  decided to use double everywhere

PR #79, opened 2017-01-09
Add sample for common Gradle project.

### LINKED ISSUE #18: Add sample gradle project to the repo

artem-zinnatullin:
Closes #18.


artem-zinnatullin:
Would be great to build samples on CI btw, @ming13?

artem-zinnatullin:
Actually, lemme try

artem-zinnatullin:
Lol, it works :) Didn't use `language: java` in `.travis.yml` because it does not support multiple languages, we'll install dependencies for other samples via `apt-get`.

arturdryomov (commented):
Actually you know what, I’m against it. I don’t really see the point. We’ll have `N` projects for `N` build systems and their flavors, store their source code, configure CI, maintain all of this and all for basically nothing. We should not care about such thing if we want to be abstract. Pre-process your Christmas photos on remote machine, if you want to, we don’t care. The real user most likely will try to integrate mainframer directly to its project and for that we will have sample ignore files, this is more than enough.

artem-zinnatullin:
Well, even though I definitely see the point of what you're talking about and have thought about it for few days before implementing this PR, I've found that without actual project it's pretty hard to write correct ignore rules, I even did a mistake in **this** Gradle sample though we use Gradle for years and `mainframer` with Gradle for more than a month.

Idea is to make samples as easy to use and understand as possible, which means — we better include not only `mainframer` configs but some build system related files (sources, etc).

Btw, most of other build systems would not require us to commit any kind of Gradle wrapper to the repo since most of them don't have such a concept, we'll commit only sources.

samples/gradle/mainframer.sh
```diff
@@ -0,0 +1,109 @@
+#!/bin/bash
```
arturdryomov: `dd`

arturdryomov (changes requested):

artem-zinnatullin:
Fixed comment, merging to add other samples

PR #91, opened 2017-01-13
Use ~/mainframer/ dir on remote machine.

### LINKED ISSUE #30: Use mainframer folder on a remote machine
This will help keeping remote home directory `~` clean and easy to navigate through in case of many projects.

artem-zinnatullin:
Closes #30.


mainframer.sh
```diff
@@ -67,7 +69,7 @@ function syncBeforeBuild {
 	echo "Sync local → remote machine..."
 	startTime=`date +%s`
 
-	COMMAND="rsync --archive --delete --relative= --compress-level=$LOCAL_COMPRESS_LEVEL "
+	COMMAND="rsync --archive --delete --rsync-path=\"mkdir -p $PROJECT_DIR_ON_REMOTE_MACHINE && rsync\" --compress-level=$LOCAL_COMPRESS_LEVEL "
```
artem-zinnatullin: `--rsync-path` is an option that allows you specify path to `rsync` executable on remote machine, trick here is to specify command that creates directory first and then invokes `rsync`, this does not require separate call to `ssh` which is good for performance, but the caveat is that not all shells support `&&`, but I hope it's not a real problem.

arturdryomov: Single quotes?

arturdryomov (changes requested):

mainframer.sh
artem-zinnatullin: Ok, I also forgot to escape `$PROJECT_DIR_ON_REMOTE_MACHINE` and it'll fail in case if we decide to add spaces.

arturdryomov (approved):

PR #98, opened 2017-01-15
Add GCC sample.

### LINKED ISSUE #58: Add sample gcc project to the repo

artem-zinnatullin:
Closes #58.


arturdryomov (approved):

samples/gcc/.mainframer/remoteignore
```diff
@@ -0,0 +1,2 @@
+# Usually you don't need to sync sources back from remote to local machine.
+*.c
```
gitanshu, line 2: also ignore *.h i guess

artem-zinnatullin: In this sample we don't have header files, but PR welcome :)

gitanshu: Will do 😬

artem-zinnatullin: Please take a look at https://github.com/github/gitignore/blob/master/C.gitignore, maybe it makes sense to include something from it.

PR #102, opened 2017-01-17
Refactor script.

arturdryomov:
Just basic renaming to make more sense.


artem-zinnatullin (changes requested):
Please do the same in `test/common.sh` for consistency.

artem-zinnatullin (approved):

PR #112, opened 2017-01-18
Exit with error code if remote command failed.

### LINKED ISSUE #110: Exit with error code if remote command errored.

artem-zinnatullin:
Closes #110.

### DIFF
--- a/mainframer.sh
+++ b/mainframer.sh
@@ -58,6 +58,7 @@ fi
 
 
 BUILD_COMMAND="$@"
+REMOTE_COMMAND_SUCCESSFUL="false"
 
 if [ -z "$BUILD_COMMAND" ]; then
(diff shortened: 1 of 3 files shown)


mainframer.sh
```diff
@@ -58,6 +58,7 @@ fi
 
 
 BUILD_COMMAND="$@"
+REMOTE_COMMAND_SUCCESSFUL="false"
```
arturdryomov, line 61: `BUILD_COMMAND` → `REMOTE_COMMAND`?

arturdryomov (changes requested):

mainframer.sh
artem-zinnatullin: in a separate PR

arturdryomov (approved):

PR #119, opened 2017-01-19
Add proper config file error messages.

arturdryomov:


artem-zinnatullin:
Would be good to write a test for that, but it's tricky, you'll need to redirect output somewhere and find required message in it.

artem-zinnatullin:
Ok, let's merge it for now since we have tests to check that config is required and add tests to check error message later.

artem-zinnatullin (approved):

PR #128, opened 2017-01-20
Print command during remote execution step.

### LINKED ISSUE #127: Print remote command during execution step.

artem-zinnatullin:
Closes #127.

Please check CI output as an example.

### DIFF
--- a/mainframer.sh
+++ b/mainframer.sh
@@ -99,7 +99,7 @@ function executeRemoteCommand {
 	startTime="$(date +%s)"
 
 	set +e
-	if ssh $REMOTE_MACHINE "echo 'set -e && cd $PROJECT_DIR_ON_REMOTE_MACHINE && echo \"$REMOTE_COMMAND\" && $REMOTE_COMMAND' | bash" ; then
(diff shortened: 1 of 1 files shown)


arturdryomov:
It produces something like this.
```
:: mainframer v2.0.0-dev

Sync local → remote machine...
Sync done: took 2 seconds.

Executing command on remote machine…

./gradlew clean build
Downloading https://services.gradle.org/distributions/gradle-3.3-bin.zip
```
I think this can be better.
```
:: mainframer v2.0.0-dev

Sync local → remote machine...
Sync done: took 2 seconds.

Executing command on remote machine…
./gradlew clean build

Downloading https://services.gradle.org/distributions/gradle-3.3-bin.zip
```

artem-zinnatullin:
Not sure. In cases when command does not output anything itself, we'll have:

```
:: mainframer v2.0.0-dev

Sync local → remote machine...
Sync done: took 2 seconds.

Executing command on remote machine…
./gradlew clean build

Execution done, took 6 seconds.
```

While with mine version it would be:

```
:: mainframer v2.0.0-dev

Sync local → remote machine...
Sync done: took 2 seconds.

Executing command on remote machine…

./gradlew clean build

Execution done, took 6 seconds.
```

Moreover, we still have `echo ""` after execution, so to keep consistency we either need to remove `echo ""` altogether or leave them both before and after command.

arturdryomov (approved):

PR #131, opened 2017-01-20
Add shellcheck as a test, fix issues it found.

artem-zinnatullin:
Closes #117.

Very nice tool!

PR #133, opened 2017-01-22
Fix $HOME expanding on client side.

artem-zinnatullin:
PR #131 introduced a bug that leads to creation of full home path of local machine user on remote machine.

Problem is that our CI does not catch this since it uses localhost with same user as remote machine.


artem-zinnatullin:
I'm merging it after CI, this is critical, don't want let somebody to download it accidentally. Code review is still appreciated and comments can be fixed in a separate PR.

PR #136, opened 2017-01-23
Change remote command printing.

arturdryomov:


mainframer.sh
```diff
@@ -96,10 +96,12 @@ function syncBeforeRemoteCommand {
 function executeRemoteCommand {
 	echo "Executing command on remote machine…"
 	echo ""
+	echo "$REMOTE_COMMAND"
```
artem-zinnatullin: This should be done on remote machine to evaluate possible env vars inside string correctly.

artem-zinnatullin (changes requested):

artem-zinnatullin (approved):

PR #139, opened 2017-01-23
Improve demo terminal output and markdown.

artem-zinnatullin:
I still think we need something more interactive than static text…

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,13 +31,13 @@ We have quite a bunch of samples showing off some practical applications.
 
 ## Demo
 
-```
+```console
 # Local execution.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -31,13 +31,13 @@ We have quite a bunch of samples showing off some practical applications.
 
 ## Demo
 
-```
+```console
```
artem-zinnatullin, line 34: Here is complete list of supported langs by GitHub's markdown https://github.com/github/linguist/blob/master/lib/linguist/languages.yml

arturdryomov (approved):

PR #140, opened 2017-01-26
Pass -t to ssh to have colored output on local machine.

artem-zinnatullin:


arturdryomov (commented):
As we discussed privately this leads to some issues, i. e. unwanted error messages in IDE and corner cases. Not sure if it worth it at this point.

artem-zinnatullin:
Ok, let's close it for now and reopen if someone will need it.

Issue #143, opened 2017-02-01
Need to add --login to bash to work with rbenv


Gazer:
Hi, I was trying maiframer but I was unable to run fastlane on the remote machine :

```
$ ./mainframer.sh fastlane help
:: mainframer v2.0.0

Sync local → remote machine...
Sync done: took 1 seconds.

Executing command on remote machine…

fastlane help

bash: line 1: fastlane: command not found

Execution failed: took 0 seconds.

Sync remote → local machine…
Sync done: took 6 seconds.

Failure: took 7 seconds.
```

After reading a lot, triying to use .bash_profile or .profile to setup rbenv correctly, I discovered that if I change the line 102 (of 2.0.0 release) to use `bash --login` it works as expected :

```
$ ./mainframer.sh fastlane help
:: mainframer v2.0.0

Sync local → remote machine...
Sync done: took 3 seconds.

Executing command on remote machine…

fastlane help

  fastlane

  CLI for 'fastlane' - The easiest way to automate beta deployments and releases for your iOS and Android apps
....
```

The .bashrc of the user I use to connecto to the remote machine ends with :

```
export PATH="$HOME/.rbenv/bin:$ANDROID_HOME/tools:$ANDROID_HOME/platform-tools:$PATH"
eval "$(rbenv init -)"
```

As I can see from the recipe to configure the remote machine there is nothing special to setup, but my ENV is different if I use or not --login. The remote system is a Ubuntu 14.04.4 LTS.

Please let me know if you need more info or some test from my side.

artem-zinnatullin:
>>The .bashrc of the user I use to connecto to the remote machine ends with 

>ends with 

That's the reason, on most Linux distributives `~/.bashrc` has ~this in the beginning:

```bash
# ~/.bashrc: executed by bash(1) for non-login shells.
# see /usr/share/doc/bash/examples/startup-files (in the package bash-doc)
# for examples

# If not running interactively, don't do anything
case $- in
    *i*) ;;
      *) return;;
esac
```

Notice `# If not running interactively, don't do anything` which basically stops evaluation of following content of the `~/.bashrc`.

For example in our setup we put required environment variables **before** this line, simply in the beginning of `~/.bashrc` so they're evaluated no matter if it's login session or not.

artem-zinnatullin:
[...]

PR #145, opened 2017-02-02
Add a note about bash initialization.

Gazer:
This is a note for #143 . English is not my first language, so I open to suggestions.

### DIFF
--- a/docs/SETUP_REMOTE.md
+++ b/docs/SETUP_REMOTE.md
@@ -33,7 +33,7 @@ If some tool require shell initialization scripts or special environment variabl
 
 Be sure to add all your configurations at the beginning of the file before this line :
 
-  ```
+  ```bash
   # If not running interactively, don't do anything
   ```
 
--- a/docs/SETUP_REMOTE.md
+++ b/docs/SETUP_REMOTE.md
@@ -29,7 +29,7 @@ There are other options like a Docker container per person though.
 
 Install tools you need to perform remote processing.
 
-If some tool require shell initialization scripts or special environment variables you should add them to your ~/.bashrc or ~/.bash_profile file depending on your remote operating system.
+If some tool require shell initialization scripts or special environment variables you should add them to your `~/.bashrc` or `~/.bash_profile` file depending on your remote operating system.
 
 Be sure to add all your configurations at the beginning of the file before this line :
 
--- a/docs/SETUP_REMOTE.md
+++ b/docs/SETUP_REMOTE.md
@@ -24,12 +24,12 @@ There are other options like a Docker container per person though.
   $ echo {SSH_KEY} >> ~/.ssh/authorized_keys
   $ chmod u+rw,go= ~/.ssh
   ```
-  
+
 ## Environment
 
 Install tools you need to perform remote processing.
 
-If some tool require shell initialization scripts or special environment variables you should add them to your `~/.bashrc` or `~/.bash_profile` file depending on your remote operating system.
+If some tool requires shell initialization scripts or special environment variables, you should add them to your `~/.bashrc` or `~/.bash_profile` file depending on your remote operating system.
 
 Be sure to add all your configurations at the beginning of the file before this line :
 
--- a/docs/SETUP_REMOTE.md
+++ b/docs/SETUP_REMOTE.md
@@ -29,6 +29,14 @@ There are other options like a Docker container per person though.
 
 Install tools you need to perform remote processing.
 
+If some tool require shell initialization scripts or special environment variables you should add them to your ~/.bashrc or ~/.bash_profile file depending on your remote operating system.
+
+Be sure to add all your configurations at the beginning of the file before this line :
+
+  ```
+  # If not running interactively, don't do anything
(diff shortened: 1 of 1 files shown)


artem-zinnatullin:
FYI: Looks like your git email is not correct, commit author has no avatar, you can change it and then  rebase commit and force push it:

```console
git config user.email "correct@email.com"
git commit --amend --reset-author
git push --force-with-lease
```

docs/SETUP_REMOTE.md
```diff
@@ -29,6 +29,14 @@ There are other options like a Docker container per person though.
 
 Install tools you need to perform remote processing.
 
+If some tool require shell initialization scripts or special environment variables you should add them to your ~/.bashrc or ~/.bash_profile file depending on your remote operating system.
```
artem-zinnatullin: `~/.bashrc`, `~/.bash_profile` https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#code

docs/SETUP_REMOTE.md
```diff
@@ -29,6 +29,14 @@ There are other options like a Docker container per person though.
...
+
+Be sure to add all your configurations at the beginning of the file before this line :
+
+  ```
```
artem-zinnatullin: ` ```bash `

artem-zinnatullin (changes requested):
Few small changes pls

Gazer:
I always forget to change the email, thanks :D

Also I've commited the other changes.

artem-zinnatullin (approved):
👍

docs/SETUP_REMOTE.md
```diff
@@ -29,6 +29,14 @@ There are other options like a Docker container per person though.
 
 Install tools you need to perform remote processing.
 
+If some tool require shell initialization scripts or special environment variables you should add them to your `~/.bashrc` or `~/.bash_profile` file depending on your remote operating system.
```
gitanshu: s/require/requires
s/variables/variables,

gitanshu (commented):
derp

docs/SETUP_REMOTE.md
artem-zinnatullin: 👍

Gazer: s/variables/variables ? should be s/variables/variable ?

docs/SETUP_REMOTE.md
```diff
@@ -29,6 +29,14 @@ There are other options like a Docker container per person though.
...
+
+Be sure to add all your configurations at the beginning of the file before this line :
+
+  ```bash
+  # If not running interactively, don't do anything
```
arturdryomov, line 37: Pretty sure it is not universal enough, some distros do not ship any default configs (i. e. Arch, Gentoo, you name it). @artem-zinnatullin, what can we do here?

artem-zinnatullin: >depending on your remote operating system

Already mentioned

>Be sure to add all your configurations at the beginning of the file before this line

Does not prevent you from adding setup you need if you don't have such line in the file, so I guess it's ok.

Also, Arch, Gentoo, etc users are usually pro-enough to find out solution on their own :)

docs/SETUP_REMOTE.md
gitanshu: variables is fine. Just put a comma after it.

docs/SETUP_REMOTE.md
arturdryomov: Basically you narrow the texting to Ubuntu users only if I understand correctly. No one else ships default configs, I don’t recall such practice.

artem-zinnatullin: Many distros do that, Amazon Linux for instance (based on Cent OS), most of Debian-like distros. Though not all of them have such a line.

Anyway I'd prefer to add this note since Ubuntu Server is very common choice.

Gazer: rbenv only mention `.bashrc` as Ubuntu specific. For other distros they recommend `.bash_profile` (https://github.com/rbenv/rbenv#basic-github-checkout).

I think that cover the common cases for new linux users. More advanced users (Arch, Gentoo, etc) may find their equivalent easily.

arturdryomov: I’m being grumpy (clearly), but I think this would be better.

```
If a build system requires Shell initialization scripts or special environment variables,
you should place them to your `~/.bashrc` or `~/.bash_profile` file depending on your remote OS.
Be careful, some Bash configurations do block config evaluation when detecting non-interactive mode.
In such case, you should place your declarations before the blocking instruction.
```

arturdryomov (changes requested):

artem-zinnatullin:
@ming13 let's merge this as is and you'll fix any nitpicks you want? :)

arturdryomov (approved):

artem-zinnatullin:
@Gazer thanks for contribution, sorry that it took so long to merge it!

PR #147, opened 2017-02-08
Fix remote_setup_script documentation

dhartwich1991:
Hey,

when trying to put this script into a file it wasn't running.
I checked the documentation and found out that the `+` needs to be after the `u` in order to make this script work.

Cheers


artem-zinnatullin (approved):
Uh, that's my bad, thanks for fixing!

dhartwich1991:
:+1: just trying to give back :) awesome project

Issue #121, opened 2017-01-19
Format time to output minutes, hours etc.


artem-zinnatullin:

dhartwich1991:
Hey, I did something in the way of this in #148

arturdryomov:
Closing this after #148 success.

artem-zinnatullin:
@dhartwich1991 thanks again for contribution!

PR #148, opened 2017-02-13
Format time to output minutes, hours and seconds

dhartwich1991:
Just to bring #121 a bit forward I implemented this little converter function to output formatted time.

In the logs it will now look like this (hh:mm:ss):

`Sync done: took 00:00:26`

@ming13 @artem-zinnatullin

### DIFF
--- a/mainframer.sh
+++ b/mainframer.sh
@@ -31,9 +31,9 @@ COMMON_IGNORE_FILE="$CONFIG_DIR/ignore"
 LOCAL_IGNORE_FILE="$CONFIG_DIR/localignore"
 REMOTE_IGNORE_FILE="$CONFIG_DIR/remoteignore"
 
- HOURS_TAG="hours"
- MINUTES_TAG="minutes"
- SECONDS_TAG="seconds"
+HOURS_TAG="hours"
+MINUTES_TAG="minutes"
+SECONDS_TAG="seconds"
 
 function read_config_property {
     grep "^${1}=" "$CONFIG_FILE" | cut -d'=' -f2
@@ -74,7 +74,7 @@ if [ -z "$REMOTE_COMMAND" ]; then
 	exit 1
 fi
 
-function convertSecondsToTime {
+function formatTime {
     ((h=${1}/3600))
     ((m=(${1}%3600)/60))
     ((s=${1}%60))
@@ -90,18 +90,9 @@ function convertSecondsToTime {
         SECONDS_TAG="second"
     fi
 
-     #See if we can emit one of the time units, because it's 0.
-     if [ "$h" -eq "0" ] && [ "$m" -eq "0" ]; then
-        printf "%d $SECONDS_TAG\n" ${s}
-     elif [ "$h" -eq "0" ]; then
-        printf "%d $MINUTES_TAG %d $SECONDS_TAG\n" ${m} ${s}
-     elif [ "$m" -eq "0" ]; then
-        printf "%d $HOURS_TAG %d $SECONDS_TAG\n" ${h} ${s}
-     elif [ "$s" -eq "0" ]; then
-        printf "%d $HOURS_TAG %d $MINUTES_TAG\n" ${h} ${m}
-     else
-        printf "%d $HOURS_TAG %d $MINUTES_TAG %d $SECONDS_TAG\n" ${h} ${m} ${s}
-      fi
+    (( $h > 0 )) && printf "%d $HOURS_TAG " ${h}
+    (( $m > 0 )) && printf "%d $MINUTES_TAG " ${m}
+    (( $s > 0 )) && printf "%d $SECONDS_TAG \n" ${s}
 }
 
 function syncBeforeRemoteCommand {
@@ -123,7 +114,7 @@ function syncBeforeRemoteCommand {
 	eval "$COMMAND"
 
 	endTime="$(date +%s)"
-	echo "Sync done: took $(convertSecondsToTime $((endTime-startTime)))"
+	echo "Sync done: took $(formatTime $((endTime-startTime)))"
 	echo ""
 }
 
@@ -144,9 +135,9 @@ function executeRemoteCommand {
 	duration="$((endTime-startTime))"
 
 	if [ "$REMOTE_COMMAND_SUCCESSFUL" == "true" ]; then
-		echo "Execution done: took $(convertSecondsToTime $duration)"
+		echo "Execution done: took $(formatTime $duration)"
 	else
-		echo "Execution failed: took $(convertSecondsToTime $duration)"
+		echo "Execution failed: took $(formatTime $duration)"
 	fi
 
 	echo ""
@@ -170,7 +161,7 @@ function syncAfterRemoteCommand {
 	eval "$COMMAND"
 
 	endTime="$(date +%s)"
-	echo "Sync done: took $(convertSecondsToTime $((endTime-startTime)))"
+	echo "Sync done: took $(formatTime $((endTime-startTime)))"
 }
 
 pushd "$PROJECT_DIR" > /dev/null
@@ -187,8 +178,8 @@ echo ""
 DURATION="$((FINISH_TIME-START_TIME))"
 
 if [ "$REMOTE_COMMAND_SUCCESSFUL" == "true" ]; then
-	echo "Success: took $(convertSecondsToTime $DURATION)"
+	echo "Success: took $(formatTime $DURATION)"
 else
-	echo "Failure: took $(convertSecondsToTime $DURATION)"
+	echo "Failure: took $(formatTime $DURATION)"
 	exit 1
 fi
--- a/mainframer.sh
+++ b/mainframer.sh
@@ -34,9 +34,6 @@ REMOTE_IGNORE_FILE="$CONFIG_DIR/remoteignore"
  HOURS_TAG="hours"
  MINUTES_TAG="minutes"
  SECONDS_TAG="seconds"
- HOURS_FORMAT="%02d"
- MINUTES_FORMAT="%02d"
- SECONDS_FORMAT="%02d"
 
 function read_config_property {
     grep "^${1}=" "$CONFIG_FILE" | cut -d'=' -f2
@@ -85,24 +82,25 @@ function convertSecondsToTime {
     #Find out the time format and correct string.
     if [ "$h" -eq "1" ]; then
         HOURS_TAG="hour"
-        HOURS_FORMAT="%01d"
     fi
     if [ "$m" -eq "1" ]; then
         MINUTES_TAG="minute"
-        MINUTES_FORMAT="%01d"
     fi
     if [ "$s" -eq "1" ]; then
         SECONDS_TAG="second"
-        SECONDS_FORMAT="%01d"
     fi
 
      #Check if hour is 0, so it doesn't need to be shown.
-     if [ "$h" -eq "0" ]; then
-        # shellcheck disable=SC2059
-        printf "$MINUTES_FORMAT $MINUTES_TAG $SECONDS_FORMAT $SECONDS_TAG\n" ${m} ${s}
+     if [ "$h" -eq "0" ] && [ "$m" -eq "0" ]; then
+        printf "%d $SECONDS_TAG\n" ${s}
+     elif [ "$h" -eq "0" ]; then
+        printf "%d $MINUTES_TAG %d $SECONDS_TAG\n" ${m} ${s}
+     elif [ "$m" -eq "0" ]; then
+        printf "%d $HOURS_TAG %d $SECONDS_TAG\n" ${h} ${s}
+     elif [ "$s" -eq "0" ]; then
+        printf "%d $HOURS_TAG %d $MINUTES_TAG\n" ${h} ${m}
      else
-        # shellcheck disable=SC2059
-        printf "$HOURS_FORMAT $HOURS_TAG $MINUTES_FORMAT $MINUTES_TAG $SECONDS_FORMAT $SECONDS_TAG\n" ${h} ${m} ${s}
+        printf "%d $HOURS_TAG %d $MINUTES_TAG %d $SECONDS_TAG\n" ${h} ${m} ${s}
       fi
 }
 
--- a/mainframer.sh
+++ b/mainframer.sh
@@ -90,9 +90,9 @@ function formatTime {
         SECONDS_TAG="second"
     fi
 
-    (( $h > 0 )) && printf "%d $HOURS_TAG " ${h}
-    (( $m > 0 )) && printf "%d $MINUTES_TAG " ${m}
-    (( $s > 0 )) && printf "%d $SECONDS_TAG \n" ${s}
+    (( h > 0 )) && printf "%d $HOURS_TAG " ${h}
+    (( m > 0 )) && printf "%d $MINUTES_TAG " ${m}
+    (( s > 0 )) && printf "%d $SECONDS_TAG \n" ${s}
 }
 
 function syncBeforeRemoteCommand {
--- a/mainframer.sh
+++ b/mainframer.sh
@@ -90,7 +90,7 @@ function convertSecondsToTime {
         SECONDS_TAG="second"
     fi
 
-     #Check if hour is 0, so it doesn't need to be shown.
+     #See if we can emit one of the time units, because it's 0.
      if [ "$h" -eq "0" ] && [ "$m" -eq "0" ]; then
         printf "%d $SECONDS_TAG\n" ${s}
      elif [ "$h" -eq "0" ]; then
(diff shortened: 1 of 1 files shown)


arturdryomov:
Thanks for this! Actually, you can achieve the same thing using `date` options.

Not sure what @artem-zinnatullin meant initially, but probably we wanted to see quote-unquote human time, i. e. `1 minute 42 seconds`.

artem-zinnatullin:
Hm, looks like it doesn't work on CI

screen shot 2017-02-14 at 11 30 44 am

>Not sure what @artem-zinnatullin meant initially, but probably we wanted to see quote-unquote human time, i. e. `1 minute 42 seconds`.

Yup, that what I wanted, sorry for not being clear enough!

@dhartwich1991 would you like to work on something like `1 minute 42 seconds` or leave it for us?

dhartwich1991:
@artem-zinnatullin hmm strange, no idea why it doesn't work on CI, but I'll figure out I guess. Yea I will take a look today to work on `1 minute 42 seconds`

mainframer.sh
```diff
@@ -89,7 +96,7 @@ function syncBeforeRemoteCommand {
 	eval "$COMMAND"
 
 	endTime="$(date +%s)"
-	echo "Sync done: took $((endTime-startTime)) seconds."
+	echo "Sync done: took $(convertsecs $((endTime-startTime)))"
```
jtietema: stupid question, but shouldn't this be `convertSecondsToTime` instead of `convertsecs`?

dhartwich1991: Of course it should :/! Will definitely provide a fix for this to make it work today :o

dhartwich1991:
Hey @artem-zinnatullin @ming13 with https://github.com/gojuno/mainframer/pull/148/commits/8f4a97295b6f5edc384eb0842c2bf05fe96e07f7 the timestamp should look now like you wanted it to :) 
- It puts `minute/minutes` etc. depending on if it's one or more time units
- Also it only shows 1 number if the number is 1
- If there are no hours we don't show them (should be enough for most cases and I hope the build doesn't take over 1 hour for anyone, but there is a safety net for this anyways)
- Example output: `Success: took 1 minute 20 seconds`

dhartwich1991:
I get an error while building :(
````
In /home/travis/build/gojuno/mainframer/test/../mainframer.sh line 101:
    printf "$MINUTES_FORMAT $MINUTES_TAG $SECONDS_FORMAT $SECONDS_TAG\n" ${m} ${s}
           ^-- SC2059: Don't use variables in the printf format string. Use printf "..%s.." "$foo".`
````

It's because I put variables in the printf to enable the dynamic formatting of the text but that is not allowed. :( Do you have any suggestions how I could achieve the same stuff?

File: mainframer.sh [hunk 2]
```diff
@@ -70,6 +77,33 @@ if [ -z "$REMOTE_COMMAND" ]; then
 	exit 1
 fi
 
+function convertSecondsToTime {
+((h=${1}/3600))
```
artem-zinnatullin: nit: please add indentation like we have in other functions (4 spaces)

mainframer.sh
```diff
@@ -70,6 +77,33 @@ if [ -z "$REMOTE_COMMAND" ]; then
...
+    MINUTES_FORMAT="%01d"
+fi
+if [ "$s" -eq "1" ]; then
+    SECONDS_TAG="second"
+    SECONDS_FORMAT="%01d"
+fi
+
+ #Check if hour is 0, so it doesn't need to be shown.
+ if [ "$h" -eq "0" ]; then
+    printf "$MINUTES_FORMAT $MINUTES_TAG $SECONDS_FORMAT $SECONDS_TAG\n" ${m} ${s}
```
artem-zinnatullin: shellcheck doesn't like it (see CI build:

```
^-- SC2059: Don't use variables in the printf format string. Use printf "..%s.." "$foo".
```

https://github.com/koalaman/shellcheck/wiki/SC2059

mainframer.sh
```diff
@@ -70,6 +77,33 @@ if [ -z "$REMOTE_COMMAND" ]; then
...
+ else
+    printf "$HOURS_FORMAT $HOURS_TAG $MINUTES_FORMAT $MINUTES_TAG $SECONDS_FORMAT $SECONDS_TAG\n" ${h} ${m} ${s}
```
artem-zinnatullin: Same as above

mainframer.sh
```diff
@@ -70,6 +77,33 @@ if [ -z "$REMOTE_COMMAND" ]; then
...
+if [ "$m" -eq "1" ]; then
+    MINUTES_TAG="minute"
+    MINUTES_FORMAT="%01d"
+fi
+if [ "$s" -eq "1" ]; then
+    SECONDS_TAG="second"
+    SECONDS_FORMAT="%01d"
+fi
+
+ #Check if hour is 0, so it doesn't need to be shown.
```
artem-zinnatullin: So it always shows minutes? Our builds usually take less than 30 seconds so we see minutes very rarely :)

artem-zinnatullin (changes requested):
Few things to fix!

mainframer.sh
dhartwich1991: Yep probably it will :D I can add another case for just showing seconds if there are no minutes, what do you think?

mainframer.sh
artem-zinnatullin: Looks like your case is valid, see exceptions: https://github.com/koalaman/shellcheck/wiki/SC2059#exceptions

dhartwich1991: Do you have any solution for this? The problem is that I need to put the Format (`%02d`) there and this is dynamic, which is why the `%s` approach didn't really work for me

artem-zinnatullin: Solution is to ignore this warning :)

dhartwich1991: Yea just did that :D

dhartwich1991:
Yea, and also the build passed 🏆 💚

mainframer.sh
```diff
@@ -31,6 +31,13 @@ COMMON_IGNORE_FILE="$CONFIG_DIR/ignore"
 LOCAL_IGNORE_FILE="$CONFIG_DIR/localignore"
 REMOTE_IGNORE_FILE="$CONFIG_DIR/remoteignore"
 
+ HOURS_TAG="hours"
+ MINUTES_TAG="minutes"
+ SECONDS_TAG="seconds"
+ HOURS_FORMAT="%02d"
+ MINUTES_FORMAT="%02d"
+ SECONDS_FORMAT="%02d"
```
artem-zinnatullin: Looks like it's not the best format, you can check logs on CI:

```
took 00 minutes 07 seconds
```

https://travis-ci.org/gojuno/mainframer/builds/201833897#L2406

mainframer.sh
```diff
@@ -70,6 +77,35 @@ if [ -z "$REMOTE_COMMAND" ]; then
...
+    if [ "$m" -eq "1" ]; then
+        MINUTES_TAG="minute"
+        MINUTES_FORMAT="%01d"
+    fi
+    if [ "$s" -eq "1" ]; then
+        SECONDS_TAG="second"
+        SECONDS_FORMAT="%01d"
+    fi
+
+     #Check if hour is 0, so it doesn't need to be shown.
```
artem-zinnatullin: Yeah, we need to omit minutes if they're not present as well as hours:

https://travis-ci.org/gojuno/mainframer/builds/201833897#L2406

artem-zinnatullin (changes requested):
Time formatting needs changes!

mainframer.sh
dhartwich1991: What would you prefer? I mean omitting minutes is clear but should seconds be just one digit in case they are 1-9?

artem-zinnatullin: Yup, I guess, all units should be just one digit if possible? cc @ming13

dhartwich1991:
Okay @artem-zinnatullin if you want you can take another look, I was also able to remove the `#shellcheck ignore` 😃

dhartwich1991:
it's also 1 digit now for values 0-9 and 0 hours and minutes are omitted :D

mainframer.sh
```diff
@@ -70,6 +74,36 @@ if [ -z "$REMOTE_COMMAND" ]; then
...
+        HOURS_TAG="hour"
+    fi
+    if [ "$m" -eq "1" ]; then
+        MINUTES_TAG="minute"
+    fi
+    if [ "$s" -eq "1" ]; then
+        SECONDS_TAG="second"
+    fi
+
+     #Check if hour is 0, so it doesn't need to be shown.
```
artem-zinnatullin: Looks like this comment is outdated now, remove?

artem-zinnatullin (approved):

mainframer.sh
dhartwich1991: Yea true, i updated it with a more explaining one

mainframer.sh
```diff
@@ -31,6 +31,10 @@ COMMON_IGNORE_FILE="$CONFIG_DIR/ignore"
 LOCAL_IGNORE_FILE="$CONFIG_DIR/localignore"
 REMOTE_IGNORE_FILE="$CONFIG_DIR/remoteignore"
 
+ HOURS_TAG="hours"
```
arturdryomov: Indentation.

mainframer.sh
```diff
@@ -70,6 +74,36 @@ if [ -z "$REMOTE_COMMAND" ]; then
 	exit 1
 fi
 
+function convertSecondsToTime {
```
arturdryomov: `formatTime`

mainframer.sh
```diff
@@ -70,6 +74,36 @@ if [ -z "$REMOTE_COMMAND" ]; then
...
+     elif [ "$h" -eq "0" ]; then
+        printf "%d $MINUTES_TAG %d $SECONDS_TAG\n" ${m} ${s}
+     elif [ "$m" -eq "0" ]; then
+        printf "%d $HOURS_TAG %d $SECONDS_TAG\n" ${h} ${s}
+     elif [ "$s" -eq "0" ]; then
+        printf "%d $HOURS_TAG %d $MINUTES_TAG\n" ${h} ${m}
+     else
+        printf "%d $HOURS_TAG %d $MINUTES_TAG %d $SECONDS_TAG\n" ${h} ${m} ${s}
+      fi
+}
```
arturdryomov: Pretty sure it is possible via [shorter and better approaches](http://unix.stackexchange.com/a/27014).

arturdryomov (changes requested):

dhartwich1991:
@ming13 Thanks for pointing me to this much cleaner solution for printing :)

mainframer.sh
```diff
@@ -31,6 +31,10 @@ COMMON_IGNORE_FILE="$CONFIG_DIR/ignore"
 LOCAL_IGNORE_FILE="$CONFIG_DIR/localignore"
 REMOTE_IGNORE_FILE="$CONFIG_DIR/remoteignore"
 
+HOURS_TAG="hours"
+MINUTES_TAG="minutes"
+SECONDS_TAG="seconds"
```
arturdryomov: These can be removed at this point.

mainframer.sh
```diff
@@ -70,6 +74,27 @@ if [ -z "$REMOTE_COMMAND" ]; then
 	exit 1
 fi
 
+function formatTime {
+    ((h=${1}/3600))
```
arturdryomov: What about `local` modifier?

mainframer.sh
```diff
@@ -70,6 +74,27 @@ if [ -z "$REMOTE_COMMAND" ]; then
...
+    ((m=(${1}%3600)/60))
+    ((s=${1}%60))
```
arturdryomov: Meaningful names are better most of the time, i. e. `hours`, `minutes`, `seconds`.

mainframer.sh
```diff
@@ -70,6 +74,27 @@ if [ -z "$REMOTE_COMMAND" ]; then
...
+
+    #Find out the time format and correct string.
```
arturdryomov: The code is trivial enough to omit comments.

mainframer.sh
```diff
@@ -70,6 +74,27 @@ if [ -z "$REMOTE_COMMAND" ]; then
...
+    if [ "$h" -eq "1" ]; then
+        HOURS_TAG="hour"
+    fi
+    if [ "$m" -eq "1" ]; then
+        MINUTES_TAG="minute"
+    fi
+    if [ "$s" -eq "1" ]; then
+        SECONDS_TAG="second"
+    fi
```
arturdryomov: Single-line?

mainframer.sh
```diff
@@ -70,6 +74,27 @@ if [ -z "$REMOTE_COMMAND" ]; then
...
+
+    (( h > 0 )) && printf "%d $HOURS_TAG " ${h}
+    (( m > 0 )) && printf "%d $MINUTES_TAG " ${m}
+    (( s > 0 )) && printf "%d $SECONDS_TAG \n" ${s}
+}
```
arturdryomov: While testing-poking around your approach I’ve made some changes and got something like this.

```bash
function formatTime {
    local time=$1

    local hours=$((time / 3600))
    local minutes=$(((time % 3600) / 60))
    local seconds=$((time % 60))

    if [ "$hours" -eq "1" ]; then HOURS_LABEL="hour"; else HOURS_LABEL="hours"; fi
    if [ "$minutes" -eq "1" ]; then MINUTES_LABEL="minute"; else MINUTES_LABEL="minutes"; fi
    if [ "$seconds" -eq "1" ]; then SECONDS_LABEL="second"; else SECONDS_LABEL="seconds"; fi

    (( hours > 0 )) && printf "%d $HOURS_LABEL " ${hours}
    (( minutes > 0 )) && printf "%d $MINUTES_LABEL " ${minutes}
    (( seconds > 0 )) && printf "%d $SECONDS_LABEL \n" ${seconds}
}
```

arturdryomov (changes requested):

mainframer.sh
dhartwich1991: Ok @ming13, got you, changed my code to your implementation mostly 👍  Thanks again for reviewing and helping me to implement this

mainframer.sh
```diff
@@ -70,6 +70,22 @@ if [ -z "$REMOTE_COMMAND" ]; then
...
 fi
 
+function formatTime {
+    local time=$1
+
+    local hours=$((time/3600))
+    local minutes=$(((time % 3600) / 60))
+    local seconds=$((time % 60))
+
+    if [ "$hours" -eq "1" ]; then HOURS_LABEL="hour"; else HOURS_LABEL="hours"; fi
```
artem-zinnatullin: Like 👍

artem-zinnatullin (approved):
Very nice!

mainframer.sh
```diff
@@ -70,6 +70,22 @@ if [ -z "$REMOTE_COMMAND" ]; then
 	exit 1
 fi
 
+function formatTime {
+    local time=$1
+
+    local hours=$((time/3600))
```
arturdryomov: `time / 3600`

mainframer.sh
```diff
@@ -70,6 +70,22 @@ if [ -z "$REMOTE_COMMAND" ]; then
...
+    local minutes=$(((time % 3600) / 60))
+    local seconds=$((time % 60))
+
+    if [ "$hours" -eq "1" ]; then HOURS_LABEL="hour"; else HOURS_LABEL="hours"; fi
+    if [ "$minutes" -eq "1" ]; then MINUTES_LABEL="minute"; else MINUTES_LABEL="minutes"; fi
+    if [ "$seconds" -eq "1" ]; then SECONDS_LABEL="second"; else SECONDS_LABEL="seconds"; fi
+
+    (( hours > 0 )) && printf "%d $HOURS_LABEL " ${hours}
+    (( minutes > 0 )) && printf "%d $MINUTES_LABEL " ${minutes}
+    (( seconds > 0 )) && printf "%d $SECONDS_LABEL \n" ${seconds}
```
arturdryomov: If you take a look at Travis CI output carefully, you can see this.

```
Sync local → remote machine...
Sync done: took 

Executing command on remote machine…
```

In other words, seconds count is `0`, so we output nothing. You add `seconds >= 0`, this should do the trick.

BTW, not sure if it is possible, but can we test this? CC @artem-zinnatullin

arturdryomov (changes requested):

mainframer.sh
dhartwich1991: I think you can not really test this since the time it needs to sync is variable, I tested some different locally with putting sleep. Like `sh mainframer.sh sleep 10` maybe a sleep 0 would give you this behaviour?

arturdryomov (approved):

arturdryomov:
Thanks and double thanks for standing up to my nitpicking!

dhartwich1991:
Thank you and for this awesome project :) actually I appreciate your nitpicking! Always happy to improve my code. Hope I can contribute more in the future 🔮

artem-zinnatullin:
@ming13 forgot to reply to your comment regarding testing:

We surely can test that, but that'll require some refactorings in order to make possible to just include (`source mainframer.sh`) functions without invoking script right away, then we can do regular unit testing.

Issue #150, opened 2017-02-18
Rename "Options" in readme


artem-zinnatullin:
During setup of Oleg Krupenich's machine, he noticed to me that he was unable to find documentation about the values in config files that were present in their repo already (`.mainframer/*`) by looking at our README and had to ask guys in his team for help.

He especially pointed that he saw `Options` but didn't realize that they're related to config files. Maybe we should rename it to `Config files` and do the same with `.md` file name?

artem-zinnatullin:
Closed by #153.

PR #152, opened 2017-02-21
Change version to 2.1.0.

arturdryomov:


arturdryomov:
# Changes for Releases page

Output time formatting is much better now, showing proper hours, minutes and seconds wording. 

Before: `62 minutes, 4 minutes, 1 minutes`. After: `1 hour 2 minutes, 4 minutes, 1 minute`.

Same pleasant numbers with better human output!

artem-zinnatullin:
@ming13 what about example(s) of old vs new time output? Just to make it more clear.

What about #150?

arturdryomov:
@artem-zinnatullin Don’t think docs are worth mentioning, users download the script and not the documentation. References are fixed, so everything should work on the web.

artem-zinnatullin (approved):

Issue #159, opened 2017-03-22
Sync on file change


bazted:
For some kinds of development like front-end, it is good to have possibility to sync files on each change. As far as I know rsync provides such functionality.
I think it will be great to have  specific configurations for file watcher.

artem-zinnatullin:
Yup, many build systems support such continuation modes, we can try to play with it. I have two concerns at the moment though:

* Slow network may result in backpressure
* `rsync` may not allow us create such a constant **two-way** sync channel (that would require some investigation and tests)

ntziolis:
Any update on this? This would significantly add usefulness to mainframer as many tools especially on the frondend provide continuous build.

artem-zinnatullin:
No updates so far, there is an unfortunate conflict between Mainframer being continuous and a build tool being continuous

Mainframer doesn't just sync your files, it also executes a command. It's unclear how should one run both and have proper execution updates

You can obviously run separate Mainframer or ssh with continous build command and then run something like `mainframer --continuous true` just to get continuous file sync, but that doesn't feel right

ntziolis:
Thank you for that! If that does continuous forth and back of provided directories than that solves our use case. Can you confirm it does both ways continous?

The toolchains that support continuous build have to deal with wrong partials all the time hence i think file sync should be just fine and we can live with the extra step of starting the build service.

artem-zinnatullin:
Wait wait, we don't have continuous execution mode in Mainframer yet. 

Let's call it `--continuous-execution` for now. It'll basically be running `mainframer remote-command` periodically/on-file change which is very different from `remote-command` being continuous itself

We're thinking about `--continuous-execution` and if we decide to go with it it'll conflict with continuous commands.

There is also an unfortunate conflict in naming with continuous file sync during execution that we're adding in 3.x (basically syncs files in parallel to the command to save time).

febs:
Well, if the synchronization of files would happen continuously then when it's build time most of the files would have been synced already, thus making the whole process even faster.

lukaville:
@febs but uploading is not a problem in most cases because developers usually make small changes in the source code and it syncs almost immediately. I think the main benefit here is that we can "pre-build" these changes remotely.

artem-zinnatullin:
Correct. 

We are also thinking about using `inotify`/`fswatch` to optimize sync for large projects, that will reduce time it takes `rsync` to understand what's changed.

PR #163, opened 2017-04-01
[Continuous Sync] Monitor and sync local file changes continuously with remote machine

balachandarlinks:
I have given it a try after checking the issue #159  @artem-zinnatullin , It would be great to know what you think about this solution. I have tested with my android project running with remote docker container. I am able to make my builds and changes are synced consistently. Even though local to remote sync is not a time-consuming task with android projects, I assume that it will be helpful for other types of projects.

1. Monitors file changes in the local project using "fswatch (http://emcrisostomo.github.io/fswatch/)" and syncs it with the remote machine.
2. fswatch monitors with a latency of 5 seconds to reduce the number of syncs initiated.
3. Do not monitor the paths specified in common ignore and local_ignore files to avoid unnecessary syncs while syncing from local machine to remote machine.

### DIFF
--- a/syncmainframer.sh
+++ b/syncmainframer.sh
@@ -1,0 +1,102 @@
+#!/bin/bash
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+
+# http://www.apache.org/licenses/LICENSE-2.0
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -e
+
+echo ":: syncmainframer v2.1.0"
+echo ""
+
+PROJECT_DIR="$(pwd)"
+PROJECT_DIR_NAME="$( basename "$PROJECT_DIR" )"
+PROJECT_DIR_ON_REMOTE_MACHINE="~/mainframer/$PROJECT_DIR_NAME"
+
+CONFIG_DIR="$PROJECT_DIR/.mainframer"
+CONFIG_FILE="$CONFIG_DIR/config"
+COMMON_IGNORE_FILE="$CONFIG_DIR/ignore"
+LOCAL_IGNORE_FILE="$CONFIG_DIR/localignore"
+
+function readConfigProperty {
+	grep "^${1}=" "$CONFIG_FILE" | cut -d'=' -f2
+}
+
+REMOTE_MACHINE_CONFIG_PROPERTY="remote_machine"
+LOCAL_COMPRESS_LEVEL_CONFIG_PROPERTY="local_compression_level"
+REMOTE_COMPRESS_LEVEL_CONFIG_PROPERTY="remote_compression_level"
+
+if [ ! -f "$CONFIG_FILE" ]; then
+	echo "Please create and fill $CONFIG_FILE."
+	exit 1
+fi
+
+REMOTE_MACHINE=$(readConfigProperty "$REMOTE_MACHINE_CONFIG_PROPERTY")
+LOCAL_COMPRESS_LEVEL=$(readConfigProperty "$LOCAL_COMPRESS_LEVEL_CONFIG_PROPERTY")
+
+if [ -z "$REMOTE_MACHINE" ]; then
+	echo "Please specify \"$REMOTE_MACHINE_CONFIG_PROPERTY\" in $CONFIG_FILE."
+	exit 1
+fi
+
+if [ -z "$LOCAL_COMPRESS_LEVEL" ]; then
+	LOCAL_COMPRESS_LEVEL=1
+fi
+
+function verifyRequiredTools {
+    if ! type fswatch &> /dev/null; then
+        echo "Please install fswatch from https://github.com/emcrisostomo/fswatch"
+        exit 1
+    fi
+}
+
+function monitorAndSyncChangesContinuously {
+	echo "Continuous sync local → remote machine..."
+    echo "Press ctrl+c to terminate."
+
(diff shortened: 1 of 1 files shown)


artem-zinnatullin:
Nice work! I imagined it as an option for `mainframer.sh`, like `mainframer.sh --continuous-sync command`.

---

Can you please check what happens in following scenario: 

1. You make change in fs
1. fswatch detects it
1. remote machine starts build
1. while build is happening on remote machine you make another change

Will it run two builds in parallel or it'll run new one only after previous finishes?

---

Another concern I have about this is that many build systems support such continuos modes  themselves (Gradle, Buck, etc), I don't see clear ways to support that with mainframer and that creates a question of "should we add support for this feature or not?"…

cc @ming13 PTAL

arturdryomov:
Looks interesting, but I have some concerns.

The desirable usecase looks something like this (I guess).

1. Run Mainframer in continuous mode.
2. Change source code.
3. The build was successful (or not?).
4. Run the application.
5. As we cannot guarantee that the continuous build actually finished we will run the build again just to be sure. In this case we have to rely on build system to support incremental builds properly.  

So the main benefit from this seems to be feeding the build system cache and incremental compilation.

BTW I have no idea how IDEA (see what I did here) will behave on constant resync with failed builds. It will try to reindex I guess and probably break resolved dependencies if they have failed.

I’m pretty sure there are similar solutions we can learn from. There is Guard for Ruby for example.

BTW 2 There is IDEA option `Make project automatically`, have no idea how it works. [Some info about that](http://stackoverflow.com/a/12744437).

balachandarlinks:
Thanks for the feedback @artem-zinnatullin and @ming13 

At present the sycmainfraner.sh script **only syncs the changes** with the remote machine as we make changes on the local machine. It does not initiate any build.

If I trigger a build after every sync then it can result in compilation errors due to incomplete changes as we cannot guarantee that the code will compile after every sync. 

Moving the continuous sync functionality inside mainframer.sh with a --continuous-sync or -cs is definitely better than a separate script. But it will not take any build parameters as it does not trigger any build.

As you guys said the continuous build is being introduced to many build systems. Gradle has the continuous build mode in an early stage (https://docs.gradle.org/current/userguide/continuous_build.html).  But when it is released, we can trigger a continuous build in the remote machine just like any other build command with mainframer.sh and it will finish and sync back only when the user kills the continuous build execution. So, we need to sync the changes from local machine to remote machine continuously for the continuous build system to work.

So the use case will be,
1. In terminal 1 user will run "$mainframer.sh --continuous-sync" which will sync the changes continuously with the remote machine and the user can press ctrl+c to stop it. 

2. In terminal 2 users will trigger a gradle build in continuous mode when it is available ($mainframer.sh ./gradlew -t testDebugUnitTes) which will make a new build whenever we sync from local machine to remote machine based on its own rules. It can take care of the compilation errors and the concerns with multiple builds based on the change introduced and the number of builds currently running. 

3. If the user runs a normal build then the local to remote sync will be immediate as the changes are already synced in advance.

What do you guys think?

arturdryomov:
My bad, I saw this as continuous build but it is a continuous sync indeed. Well, in that case we can definitely introduce a `--sync` or `--sync-daemon` option to push changes to remote machine all the time. @artem-zinnatullin, any thoughts on this?

balachandarlinks:
@ming13 No worries :) I would prefer to call it `--continuous-sync` ( `-c` in short form ). IMHO `--sync` refers one-time sync and our script is going to be a foreground process and not a daemon. What do you think?

balachandarlinks:
@artem-zinnatullin @ming13 Hey folks, If you guys can give me a go ahead with moving this functionality inside mainframer.sh then I can make the additional changes. I am just waiting for your green flag :)

artem-zinnatullin:
Sorry for long pause @balachandarlinks, 

So we've discussed this thing in details with @ming13 and I've tried `./gradlew --continuous` to find out how it could fit into this, what we've decided for now:

1. We're ~ok with adding such mode to `mainframer.sh` if we'll be able to solve most of the corner cases.
2. Mode should allow both `local-to-remote` and `two-way` syncs with `local-to-remote` enabled by default. `two-way` sync is required for continous builds provided by build system, so you could use results on your local machine (ie web development), `two-way` sync would require `fswatch` on remote machine as well. 
3. Mode should allow user to pass command that'll be invoked on remote machine after each sync. If you don't need it you can just pass `echo noop` or we can allow invocations with empty command.
4. Mode should be added as a parameter and not affect execution nor performance of old use cases and `mainframer.sh` should not require `fswatch` for invocations without this mode.

This is what we want to target, of course we can implement it step by step, ie add `two-way` sync later.

@balachandarlinks feel free to comment! 

@ming13 PTAL

balachandarlinks:
@artem-zinnatullin  Sounds like a good plan 👍 We can investigate further on case by case.

So let's have a roadmap,
1. Continuous local to remote sync mode (--continuous-sync) without remote command execution​ support.
2. Adding continuous remote to local sync to --continuous-sync mode.
3. Adding remote command execution support to continuous sync mode.
I agree with your thoughts on fswatch requirement and supporting existing usecase as it is. 

While doing my code changes, I thought that it would be good to follow a common code style/standard guideline.  Can we create one or accept a widely used one. There is a Google code guideline for bash scripts. It was looking good. Please take a look and let your what you guys think.

Google style guideline : https://google.github.io/styleguide/shell.xml

Note: I am currently in vacation and will resume only by the first week of may. So I may not be able to update my PR immediately. Apologies :(

artem-zinnatullin:
>I thought that it would be good to follow a common code style/standard guideline. Can we create one or accept a widely used one.

Good point! Created separate issue for codestyle #165. Since you're on vacation this gives us some time to apply codestyle first so your changes would not affect whole file, then we can start accepting small prs to main or separate feature branch for `--continuos-sync`and see how it fits the project and solves real world use cases :)

@balachandarlinks have a good rest on your vacation and don't write any line of code! At least not in our projects haha 🏖

balachandarlinks:
@artem-zinnatullin My sincere apologies for keeping it open for a long time without progress. I feel guilty :/ I will close this PR and will open a new one soon.

arturdryomov:
@balachandarlinks, don’t be, everything is fine ;-)

PR #171, opened 2017-05-19
Use same dots symbols everywhere.

arturdryomov:


dmitry-novikov (approved):

artem-zinnatullin (changes requested):
I've made real dots for a reason, not all terminal/OS combinations can display... properly :(

arturdryomov:
@artem-zinnatullin, so which dots should we use everywhere? Regular or Unicode ones?

artem-zinnatullin:
Regular I think, for instance (it's not the case with Mainframer though)
Jenkins can't display unicode variant properly and I have to use regular
dots in all CI related projects

arturdryomov:
Done.

artem-zinnatullin (approved):

PR #181, opened 2017-11-23
Add integration for AWS EC2

crysxd:
We used a custom version of mainframer 1.1.0 since a couple of months. I integrated our changes into the main project. 

Mainframer can now use a AWS EC2 instance as remote build machine and is able to start the instance automatically if the instance is stopped.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 Tool that allows you to move build process from a local machine to a remote one.
 
-Remote machine ought to be much faster than a laptop. 
+Remote machine ought to be much faster than a laptop.
 With `mainframer` you can free up your local machine for better things —
 like editing source code in your IDE without lags and
 freezes, being able to actually _use_ your computer when the build is happening.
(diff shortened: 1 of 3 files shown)


crysxd:
I'll fix the test issues...

crysxd:
All test complete on my local machine but Travis says `The log length has exceeded the limit of 4 MB (this usually means that the test suite is raising the same exception over and over). The job has been terminated`. Do you know how to solve this issue?

artem-zinnatullin:
Hi @crysxd!

I was about to shut down your PR and suggest to maintain a separate fork that we can list in the README because we want to keep Mainframer itself as a generic tool that doesn't know about 3rd-party services like AWS EC2

BUT, I think we have a better option - plugins system! Please check issue #182, it should allow us to keep Mainframer as a generic tool but at the same time enable integrations like this without pain.

So I'm closing this PR, please comment in #182 if you have some questions/ideas, and keep an eye on the development of plugin system, we'll be happy if you could help us test its usability before we release an update of Mainframer with it!

Sorry about that and thanks for contribution, someone can actually copy the code from your PR to use the integration in the meantime :)

PR #184, opened 2017-12-29
Add config option to display local project dir in output

tony-root:
The actual command run on remote may sometimes contain absolute references to files, compiler warnings and errors being the most common case. 
These references are often clickable (in IDEA for example), but as remote project directory does not match the local one, a click fails to resolve the actual file.

The introduced config option `local_dir_in_output` will, when enabled, replace all occurrences of the REMOTE_PROJECT_DIR to LOCAL_PROJECT_DIR. This will make the file links in the output clickable.

@artem-zinnatullin @ming13 , PTAL, if you are OK with the idea, I will than update README as well.

### DIFF
--- a/mainframer
+++ b/mainframer
@@ -38,6 +38,7 @@ function readConfigProperty {
 REMOTE_MACHINE_CONFIG_PROPERTY="remote_machine"
 LOCAL_COMPRESS_LEVEL_CONFIG_PROPERTY="local_compression_level"
 REMOTE_COMPRESS_LEVEL_CONFIG_PROPERTY="remote_compression_level"
+LOCAL_DIR_IN_OUTPUT_CONFIG_PROPERTY="local_dir_in_output"
 
 if [ ! -f "$CONFIG_FILE" ]; then
 	echo "Please create and fill $CONFIG_FILE."
@@ -47,6 +48,7 @@ fi
 REMOTE_MACHINE=$(readConfigProperty "$REMOTE_MACHINE_CONFIG_PROPERTY")
 LOCAL_COMPRESS_LEVEL=$(readConfigProperty "$LOCAL_COMPRESS_LEVEL_CONFIG_PROPERTY")
 REMOTE_COMPRESS_LEVEL=$(readConfigProperty "$REMOTE_COMPRESS_LEVEL_CONFIG_PROPERTY")
+LOCAL_DIR_IN_OUTPUT=$(readConfigProperty "$LOCAL_DIR_IN_OUTPUT_CONFIG_PROPERTY")
 
 if [ -z "$REMOTE_MACHINE" ]; then
 	echo "Please specify \"$REMOTE_MACHINE_CONFIG_PROPERTY\" in $CONFIG_FILE."
@@ -61,6 +63,9 @@ if [ -z "$REMOTE_COMPRESS_LEVEL" ]; then
 	REMOTE_COMPRESS_LEVEL=1
 fi
 
+if [ -z "$LOCAL_DIR_IN_OUTPUT" ]; then
+	LOCAL_DIR_IN_OUTPUT="false"
+fi
 
 REMOTE_COMMAND="$*"
 REMOTE_COMMAND_SUCCESSFUL="false"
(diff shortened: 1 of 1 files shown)


mainframer
```diff
@@ -115,8 +120,18 @@ function executeRemoteCommand {
...
 
 	set +e
-	if ssh "$REMOTE_MACHINE" "echo 'set -e && cd '$PROJECT_DIR_ON_REMOTE_MACHINE' && echo \"$REMOTE_COMMAND\" && echo "" && $REMOTE_COMMAND' | bash" ; then
-		REMOTE_COMMAND_SUCCESSFUL="true"
+	if [ "$LOCAL_DIR_IN_OUTPUT" == "true" ]; then
+		REMOTE_PROJECT_DIR_ABSOLUTE=$(ssh "$REMOTE_MACHINE" "cd $PROJECT_DIR_ON_REMOTE_MACHINE && pwd")
+
+		if ssh "$REMOTE_MACHINE" "echo 'set -e && cd '$PROJECT_DIR_ON_REMOTE_MACHINE' && echo \"$REMOTE_COMMAND\" && echo "" && $REMOTE_COMMAND' | bash" \
+		2> >(sed -l "s|$REMOTE_PROJECT_DIR_ABSOLUTE|$PROJECT_DIR|g" >&2) \
+		1> >(sed -l "s|$REMOTE_PROJECT_DIR_ABSOLUTE|$PROJECT_DIR|g" >&1); then
```
tony-root, line 130: Both stdin and stderr are processed without redirecting stderr to stdin to preserve the stderr coloring.

tony-root, line 130: -l is important to process the output line-by-line instead of waiting for the remote build to complete

dmitry-novikov (approved):
Very nice!

tony-root:
Hi @artem-zinnatullin, @ming13,

Looks like the Travis CI build process generates over 4M logs and gets shut down because of it. Do you have an idea on where to raise the 4M logs? Any other suggestions?

https://travis-ci.org/gojuno/mainframer/builds/322881594

```
...
Info: Preparing "Install Android SDK Platform 25 (revision: 3)".
[====================                   ] 51% Unzipping... android-7.1.1/data/la
The log length has exceeded the limit of 4 MB (this usually means that the test suite is raising the same exception over and over).
The job has been terminated
```

artem-zinnatullin:
@AntonRutkevich there is no good workaround for 4mb limit on Travis other than reducing amount of logs we produce during build :( (faced this in StorIO)

Btw, if your use case happens in IntelliJ - [Mainframer IntelliJ Plugin](https://github.com/elpassion/mainframer-intellij-plugin) handles path mapping, do you use it?

tony-root:
@artem-zinnatullin 

I am indeed using the Mainframer IntelliJ plugin.
What kind of path mapping are you referring to? What I'm trying to achieve is to be able to click the links in the compiler warnings/erros output inside IntelliJ. I could not find such setting in the Mainframer plugin readme, settings etc.

tony-root:
Also, regarding the build output: as you see it's not something introduced by this PR, it's about environment setup, so I assume that no PR in this repo will be able to build successfully..

artem-zinnatullin:
I'll be checking Mainframer CI soon

Here are few issues in Mainframer IntelliJ Plugin regarding it:

- https://github.com/elpassion/mainframer-intellij-plugin/issues/12
- https://github.com/elpassion/mainframer-intellij-plugin/issues/162

I don't have time to check plugin sources, but it might be that they only apply mapping in case of error, better double check with them and open an issue there if that's the case :)

tony-root:
@artem-zinnatullin 
Thanks, I'll check with them. Although I'm still not sure if it's better to keep it as a part of plugin or mainframer itself. 

Have a good weekend :)

artem-zinnatullin:
Yep, I agree, this might be a great addition to Mainframer itself

Have a good weekend man!

artem-zinnatullin:
@AntonRutkevich wanna rewrite this in Rust :trollface:?

If you could add integration tests in this PR, I'd be happy to merge so later I could reuse them for Rust implementation :)

Issue #185, opened 2018-02-05
OS-wide usage


arturdryomov:
After using Mainframer for a while with different projects we’ve been pointed to a specific pain point — shipping the tool with _a_ project forces you to ship it with _all_ projects.

In other words, the suggested way is to include the `mainframer.sh` in a VCS of choice and update it manually on a project-level basis. Unfortunately it does not scale well.

I think it is time to evolve Mainframer to a system-wide tool, so it can be _just like big boys_.

#### TODO

- [x] Rename `mainframer.sh` to `mainframer`.
- [ ] Support global configuration file in `$XDG_CONFIG_HOME` and `$HOME`, respecting local overrides. `ignore`, `localignore`, `remoteignore` will stay on a project level.
- [ ] Provide a Homebrew Tap since I doubt we can get into the Homebrew Core. We can try though. There are too many Linux distributions and package formats to support all (or any) of them, so I would leave Linux out of scope at this point.
- [ ] Change naming from `mainframer` to Mainframer everywhere to provide a better _brand_ experience.
- [ ] Change description from _remote builder_ to _sync and remote execution_ or something like that.

arturdryomov:
Homebrew Formula is mostly ready and actually works.

```ruby
class Mainframer < Formula
  desc "Sync and remote execution"
  homepage "https://github.com/gojuno/mainframer"

  version "2.1.0"
  url "https://github.com/gojuno/mainframer/archive/v#{version}.tar.gz"
  sha256 "78f9928527720eeeda4997f048591518fdbf76c8b894bf3c827363a8a6d12d30"

  def install
    bin.install "mainframer.sh"
  end
end
```

artem-zinnatullin:
Do we want to finish initial 3.0 before going to repositories like Homebrew?

arturdryomov:
What do you have in mind for 3.0?

artem-zinnatullin:
Everything from your list is blocking and I also want that, other than that:

- #182 Plugin system (optional, shouldn't block 3.0 release)

PR #186, opened 2018-02-08
added a default configuration file

Pozzoooo:
Thats really useful for my scenario, I work with a lot of different Android projects, and it is quite boring to configure mainframer to all of them.
This also might be beneficial when having different mainframer configuration between devs on the same project.

The only issue that I found so far is that sometimes I run mainframer in the wrong dir, and then it starts synchronise the full folder hehehe.


artem-zinnatullin:
This is definitely something we want, but in a different manner. We need to carefully merge some of the properties between config declared in `$HOME` and project folder and prioritize settings from project folder.

Please keep using workaround from this PR in your setup (since you have own copy of Mainframer), while we'll work on proper global configuration support, this is gonna be part of big task to make Mainframer OS-wide tool #185

Issue #182, opened 2017-11-24
Plugins system!


artem-zinnatullin:
At the moment the only ways to customize Mainframer behavior are:

- Modify the source code of Mainframer (easy, since we ship it as a bash script, but it's easy to break compatibility and it complicates updates for user)
- Make a custom ssh server that'll do necessary work once Mainframer connects to it (pretty hard, limited functionality,  but doesn't break compatibility and Mainframer can be updated without problems)

There are multiple users (including myself) who want custom behavior like @AndreasBackx with Docker integration #167 and @crysxd with AWS EC2 integration #181.

We can of course suggest people to fork Mainframer and maintain own fork with the integration (which complicates life for everyone).

Or we can try to design and develop plugins/hooks system that'll add extension points to Mainframer so such integrations could be maintained as separate projects (that we can list in README) and plugged into Mainframer by modifying only config files.

So far I see it as config entries that'll allow user to specify plugins (in form of executable binaries/scripts) to be run:

- before sync (local machine)
- before remote command execution (local machine)
- before remote command execution (remote machine)
- after remote command execution (remote machine)
- after remote command execution (local machine)
- after sync (local machine)

Additional information (like working directory, user command, etc) can be passed as environment variables, plus plugins are of course welcome to read Mainframer configs and store their configs in `.mainframer` folder (we'll just need to agree on file names to avoid conflicts).

@ming13 @dmitry-novikov @yunikkk @ghostbuster91, @pawegio thoughts?

artem-zinnatullin:
Most likely we'll need to allow plugins to resolve some configuration parameters for Mainframer, ie dynamically resolve remote machine connection parameters.

artem-zinnatullin:
I still think this is very important feature, but I've removed it from 3.0.0 milestone. 

We can ship initial 3.x without it and add it later.

PR #187, opened 2018-02-13
Add link to Mirakle

skrugly:
Apart from current Mainframer functionality [Mirakle](https://github.com/Instamotor-Labs/mirakle/) is highly extendable thanks to Gradle plugin system and supports remote project dir replacement described [here](https://github.com/gojuno/mainframer/pull/184).

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,7 @@ freezes, being able to actually _use_ your computer when the build is happening.
 - [Setup](#setup)
 - [Demo](#demo)
 - [3rd-party IntelliJ Mainframer Plugin](#3rd-party-intellij-mainframer-plugin)
+- [3rd-party Mainframer analogue for Gradle](#3rd-party-mainframer-analogue-for-gradle)
 - [How to Run Tests (for contributors)](#how-to-run-tests-for-contributors)
 - [License](#license)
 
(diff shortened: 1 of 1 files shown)


artem-zinnatullin:
Hey @Adambl4,

We definitely appreciate more native integration of Mainframer with Gradle, the only problem atm is that we'd like to keep Mainframer itself as both a standard and a reference implementation for couple reasons:

- It allows us add new features without need in repeating them in other implementations
- It keeps things more consistent in the community (here I'm referencing to custom Gradle DSL that Miracle provides)

Mirakle implements Mainframer standard in a way that you support Mainframer config files which is really great :+1:, however as we add more features like #188 or #182 Mirakle will have to basically duplicate them internally which might result in different behavior, bugs and other ways of inconsistency (+ it's just more work for you)

It would be super great if you could call Mainframer instead of ssh/rsync in Mirakle so it would implement both native Gradle integration as well as all the features we bundle into Mainframer itself! (you could add a parameter for path of Mainframer executable)

What do you think?

---

There is a small chance that because of deeper integration with system like Gradle you'll be able to make sync more effective, but as I was thinking about it for quite some time I wasn't able to convince myself that it'll be the case in real life

PR #189, opened 2018-02-19
global config with merge

Pozzoooo:
As suggested on my last PR, now I've used a different approach.

- We have a global configuration folder inside user $HOME/.mainframer;
- Each configuration will first be taken from local .maframer folder, and if it is not set there, then and only then it will the global configuration;
- Ignore files are not going to be merged, it will either use the local one, or use the global if it does not exists;
- If a directory does not contain a local .mainframer with a config file folder, then mainframe will not run, but it will run even if the config file is empty when all configs are available on global config.

Known issues:
- User still able to run mainframer inside $HOME dir, as it has its "local" .mainframer folder, not sure if we should add a check for it;

I hope it makes more sense.


mainframer
```diff
@@ -134,6 +157,25 @@ function executeRemoteCommand {
 	echo ""
 }
 
+function sendServerCommand {
```
artem-zinnatullin, line 161: ??

Pozzoooo: This should not be here... I was testing some alternative to pass on SIGINT without interactive ssh, gonna remove it.

Pozzoooo:
I've just seen the new PR to rewrite the tool in Rust, so there is probably not much value on keeping this PR.

artem-zinnatullin:
Well, we'll write this functionality in Rust yes, but I still wanted to merge this PR if you could add integration tests, because we can reuse them :)

artem-zinnatullin:
@Pozzoooo so what do you think about adding integration tests for this so we could merge it? :)

Pozzoooo:
Sounds good, gonna try to do it over the weekend

artem-zinnatullin:
Awesome! Even if we merge Rust PR first, I'd just rebase this one and pick up the tests

To run tests you just need Docker and then you can run them by executing `ci/build.sh` :)

Pozzoooo:
I've sent some work in progress, couldn't finish it all over the weekend... my idea is to replicate that for all the global's ignore and config files and then add some more specific ones to make sure the locals are overridden the global.

Pozzoooo:
@artem-zinnatullin 
Sorry for the long delay, but there it goes.
I have created tests for the 3 ignore files and the config file.

artem-zinnatullin:
Quick update:

I've created separate issue to track and discuss this feature #213, specifically I want to make sure we agree with folder name and behavior.

I'm still planning to merge this PR (I'll only pick integration tests as code change is now out of sync with Rust version)

No action required from you :)

Pozzoooo:
Probably too old by now, and I think I messed up with the branch I used to create the PR

artem-zinnatullin:
Don't worry, would be great if you could review the PR with this for 3.x once we start it :)

Issue #190, opened 2018-02-23
Travis CI: log length has exceeded the limit of 4 MB


artem-zinnatullin:
I think they've changed terminal session parameters, log seems to be overflooded by repeating interactive messages (like progress output) even from things like git clone which Travis does on its own.

Example: https://api.travis-ci.org/v3/job/340900769/log.txt

However we still should reduce output of things like `apt-get`, Android SDK installation and so on.

artem-zinnatullin:
Closed by #191.

Issue #188, opened 2018-02-14
Continuous sync from remote to local machine during remote command execution.


artem-zinnatullin:
Currently Mainframer has 3 main steps that run in serialized manner one after another:

1. sync from local to remote machine
1. remote command execution
1. sync from remote to local machine

Step №1 is typically fast enough and not much we can do here, optimization of step №2 is not something we can do in Mainframer, but step №3 actually looks like something we can optimize in Mainframer.

Instead of waiting for remote command to finish, we can try to run step №3 in parallel to step №2, that can significantly reduce overall execution time.

Quick lookup suggests that `rsync` doesn't support "continuous" sync. 
We can try hooking into filesystem changes on remote machine and trigger rsync. Although I feel like that would require 2 additional ssh connections because remote machine can't directly ssh into local machine.

@ming13 thoughts?

arturdryomov:
It might work, but there is an issue regarding a _black box_ nature of the execution phase. The execution itself _might_ produce some intermediate files that are being discarded on completion. In the suggested scenario we are going to copy them anyway, i. e. introducing redundancy of unknown amounts (since we don’t know if it is going to be GiB or KiB). Plus an IDE or an another client-side indexing thingy potentially can go nuts trying to reindex potentially not valid state (since the copy operation is not sync, i. e. intermediate files can be out of order).

artem-zinnatullin:
These are valid concerns for sure, but I think it still might worth it.

Unnecessary intermediate files most likely can be ignored through `remoteignore` config. 

Afaik IDEs (IntelliJ) don't hook into filesystem notifications, so if you keep using Mainframer IntelliJ Plugin it will only trigger IDE file sync after Mainframer exit which we will do only after final `rsync` run after remote command execution.

artem-zinnatullin:
```console
rsync warning: some files vanished before 
they could be transferred (code 24) at main.c(1679) [generator=3.1.3]
```

Something I'll have to deal with…

artem-zinnatullin:
I guess warning is fine and actually points to intermediate files that should be ignored via config, but this one is error:

```console
Sync remote → local machine failed: rsync exit code '24',
rsync stdout 'cannot delete non-empty directory: buck-out'
```

(Noice error output btw, 😽myself)

artem-zinnatullin:
mhm, this problems seems to be consistently appearing during clean tasks like 

```console
mainframer ./gradlew clean
```

Doesn't happen during clean/incremental builds so far.

[...]

PR #191, opened 2018-02-25
Rewrite Mainframer in Rust.

artem-zinnatullin:
Sooo, yeah lol.

General reasons to abandon Bash:

- Upcoming "Continuous sync" feature #188 that should significantly speed up most of workflows needs proper multithreading which is hard to do with Bash
- Bash makes simple things hard, like substrings, collections, etc
- Bash lacks simple programming apis which force us to use other tools like `grep` which work differently on GNU (Linux) and BSD (macOS) and drive us crazy
- Right from start we knew this day will come, it was just a matter of time, Bash served us well till this day though

Ok, but why Rust:

- Rust is a cross-platform system language
- Rust is stable (1.x), sorry Kotlin Native
- Rust doesn't have GC - less runtime overhead, sorry Go
- Rust in general is super fast and follows "Zero-cost abstractions" idea, however Mainframer bottlenecks on network and remote command execution
- Rust is memory safe (with pretty safe multithreading as a side-effect)
- Rust is my personal hipster choice for 2018, sorry Kotlin (Native too) you're not a hipster choice anymore after 2017 Google announcement
- Rust has very good set of design decisions and a compiler that tries very hard to catch errors before runtime and explain them so you don't have to google
- Rust has [very well written documentation](https://doc.rust-lang.org/book/second-edition/ch01-00-introduction.html)

---

Interesting side-effects:

- We now have not only Integration tests (which Rust version passes with breaze), but also Unit tests! yay
- We can now add much more complicated features like #188 mentioned above, parameters parsing, file parsing and whatever else we might need in future
- Mainframer is now 2x faster than version written in Bash on a simple benchmark `mainframer echo yo` with no files to sync and localhost as remote machine. In concrete numbers: `0.02` - `0.03` seconds instead of `0.05` - `0.06` seconds and `6%` CPU use instead of `9%` which is just *nice* to have

---

How to catch-up with this PR and reason about Rust code:

- Rust Documentation https://doc.rust-lang.org/book/second-edition/ch01-00-introduction.html
- Rust by Example https://rustbyexample.com/index.html
- `cargo build` to build debug version (produces `target/debug/mainframer`)
- `cargo build --release` to build release version (produces `target/release/mainframer`)
- `cargo test` to run unit tests
- `ci/build.sh` to run everything we run on CI (you just need Docker)
- I use IntelliJ with Rust plugin for development, but you can use whatever editor you like

Edit: grammar, add a note about IDE


artem-zinnatullin:
This PR also fixes #190 log limit on Travis as I was running tests too often and it just annoyed the hell out of me.

Cargo.toml
```diff
@@ -0,0 +1,6 @@
+[package]
+name = "mainframer"
+version = "2.1.0"
```
artem-zinnatullin, line 3: In future I'd like to detach version from source code and get it from current git tag, but we didn't do it with Bash so for simplicity I didn't do it in this PR

ci/docker/Dockerfile
```diff
@@ -22,14 +22,18 @@ ENV ANDROID_HOME /opt/android-sdk-linux
...
-    eval $ANDROID_SDK_INSTALL_COMPONENT '"tools"' && \
-    eval $ANDROID_SDK_INSTALL_COMPONENT '"platform-tools"' && \
-    eval $ANDROID_SDK_INSTALL_COMPONENT '"build-tools;25.0.2"' && \
-    eval $ANDROID_SDK_INSTALL_COMPONENT '"platforms;android-25"'
+    echo "Installing Android SDK components" && \
+    eval $ANDROID_SDK_INSTALL_COMPONENT '"tools"' > /dev/null && \
+    eval $ANDROID_SDK_INSTALL_COMPONENT '"platform-tools"' > /dev/null && \
+    eval $ANDROID_SDK_INSTALL_COMPONENT '"build-tools;25.0.2"' > /dev/null && \
+    eval $ANDROID_SDK_INSTALL_COMPONENT '"platforms;android-25"' > /dev/null && \
+    chmod -R 777 "$ANDROID_HOME"
```
artem-zinnatullin: Moved it here, previously we gave ownership of this folder to the build user in entrypoint script which was slowing down iterative testing as it was running every time on container startup

src/ignore.rs
```diff
@@ -0,0 +1,36 @@
...
+
+        let mut local_ignore_file = working_dir.to_path_buf();
+        local_ignore_file.push(".mainframer/localignore");
+
+        let mut remote_ignore_file = working_dir.to_path_buf();
+        remote_ignore_file.push(".mainframer/remoteignore");
+
+        Ignore {
+            common_ignore_file: match common_ignore_file.exists() {
+                true => Some(common_ignore_file.to_path_buf()),
```
artem-zinnatullin, line 23: I do pattern-matching for booleans in all languages :trollface:

src/args.rs
```diff
@@ -0,0 +1,47 @@
...
+
+impl Args {
+    pub fn parse(raw_args: Vec<String>) -> Result<Args, String> {
+        match raw_args.len() {
+            0 => Err(String::from("Please pass remote command.")), // TODO more user friendly message, for now it's consistent with Bash version.
+            _ => Ok(Args {
+                command: {
+                    let str: String = raw_args
+                        .iter()
+                        .cloned()
```
pepyakin: Because you own `raw_args` you can avoid cloning by calling `.into_iter()`. This will consume `raw_args` and turn it into an iterator that yields owned Strings.

src/args.rs
```diff
@@ -0,0 +1,47 @@
...
+                        // Add space between parameters.
+                        .map(|arg| format!("{} ", arg))
```
pepyakin: I guess, this isn't important, but this code will allocate a new String for result. This is sorta wasteful given that you already have a owned `String` that you can modify here.

src/args.rs
```diff
@@ -0,0 +1,47 @@
...
+                        .collect();
```
pepyakin: ... but better, I guess you are looking for something like Kotlin's `joinToString`. Rust also [have similar](https://doc.rust-lang.org/std/slice/trait.SliceConcatExt.html#tymethod.join) method.
It is unfortunate this function is not so discoverable.

src/time.rs
```diff
@@ -0,0 +1,141 @@
+use std::time::Duration;
+
+pub fn format_duration(duration: &Duration) -> String {
```
pepyakin: I guess, you don't actually want to pass `Duration` by a value since it is a `Copy` type.

src/time.rs
```diff
@@ -0,0 +1,141 @@
...
+
+    let seconds_label = match seconds {
+        1 => "second",
+        _ => "seconds"
+    };
+
+    let mut text = String::new();
+
+    if hours > 0 {
+        text.push_str(&format!("{value} {label}", value = hours, label = hours_label));
```
pepyakin: If you import `std::fmt::Write` then you will be able to use a `write!` macro for formatting strings directly into a `text` without an extra String being allocated.

src/sync.rs
```diff
@@ -0,0 +1,103 @@
...
+}
+
+fn project_dir_on_remote_machine(local_project_dir_name: &str) -> String {
+    format!("~/mainframer/{}", local_project_dir_name)
+}
+
+fn apply_exclude_from(rsync_command: &mut Command, exclude_file: &Option<PathBuf>) {
+    match exclude_file.clone() {
+        Some(value) => {
+            rsync_command.arg(format!("--exclude-from={}", value.to_string_lossy()));
```
pepyakin: Can you use something like?

```rust
match exclude_file {
    Some(ref value) => {
         // ...
         value.to_string_lossy()
    }
}
```

src/args.rs
```diff
@@ -0,0 +1,47 @@
+#[derive(Debug, PartialEq, Eq)]
+pub struct Args {
+    pub command: String
+}
+
+impl Args {
+    pub fn parse(raw_args: Vec<String>) -> Result<Args, String> {
+        match raw_args.len() {
+            0 => Err(String::from("Please pass remote command.")), // TODO more user friendly message, for now it's consistent with Bash version.
```
pepyakin, line 9: FWIW, `String::from(x)` is equivalent to `x.into()`.

artem-zinnatullin:
@pepyakin big thanks for review! Really appreciate it, will address your comments

I've definitely haven't gotten all quirks of reference and standard type system yet. I believe I can make function parameters more generic in terms of ownership on the data passed to a function and definitely need to learn more ways of using standard functions and types

Will be happy for another round of review from you once I fix already posted feedback!

src/time.rs
```diff
@@ -0,0 +1,141 @@
...
+    text
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn format_duration_0_seconds() {
+        assert_eq!(format_duration(&Duration::from_secs(0)), String::from("0 seconds"));
```
arturdryomov: Seems like you can move the assertion to a function named `assert_format` which will have `seconds` and `expectedText` as arguments.

src/time.rs
```diff
@@ -0,0 +1,141 @@
+use std::time::Duration;
+
+pub fn format_duration(duration: &Duration) -> String {
+    let raw_seconds = duration.as_secs();
```
arturdryomov: `total_seconds`

test/test.sh
```diff
@@ -20,11 +20,22 @@ function printTestResults {
 # Hook to exit happened either because of success or error.
 trap printTestResults EXIT
 
-echo "Checking for syntax errors…"
-bash -n "$DIR/../mainframer"
+pushd "$DIR/../" > /dev/null
+
+echo "Building debug version of Mainframer..."
+cargo build
```
arturdryomov, line 26: Is it a good idea to include some additional debug flags, like `-W` or something? Maybe even use [Clippy](https://github.com/rust-lang-nursery/rust-clippy)?

src/time.rs
```diff
@@ -0,0 +1,141 @@
...
+        }
+
+        text.push_str(&format!("{value} {label}", value = minutes, label = minutes_label));
+    }
+
+    if !text.is_empty() {
+        text.push(' ');
+    }
+
+    text.push_str(&format!("{value} {label}", value = seconds, label = seconds_label));
```
arturdryomov: The formatting `{value} {labels}` asks for moving to a separate function.

Instead of doing `String` concat yourself you’ll be better off using [`join`](https://doc.rust-lang.org/std/slice/trait.SliceConcatExt.html#tymethod.join). In Kotlin it will be `listOf("hours", "minutes").joinToString(separator = "  ")`.

src/time.rs
```diff
@@ -0,0 +1,141 @@
...
+
+    text
+}
+
+#[cfg(test)]
+mod tests {
```
arturdryomov: Is it a common practice in Rust to have tests in the same file as the main source set?

src/sync.rs
```diff
@@ -0,0 +1,103 @@
+use config::Config;
+use ignore::Ignore;
+use std::path::PathBuf;
+use std::process::Command;
+
+// TODO add internal version of sync functions with closures as parameters to unit test properly.
+pub fn sync_local_to_remote(local_project_dir_name: &str, config: &Config, ignore: &Ignore) -> Result<(), String> {
```
arturdryomov, line 7: `sync_remote_to_local` and `sync_local_to_remote` seem awfully similar. Most likely they can be unified somehow. At least the `result` part is the same.

src/sync.rs
```diff
@@ -0,0 +1,103 @@
...
+    let mut command = Command::new("rsync");
+
+    command
+        .arg("--archive")
+        .arg("--delete")
+        // Create (if not exists) project dir on remote machine.
+        .arg(format!("--rsync-path=mkdir -p {} && rsync", project_dir_on_remote_machine(local_project_dir_name)))
+        .arg(format!("--compress-level={}", config.local_compression_level));
+
+    apply_exclude_from(&mut command, &ignore.common_ignore_file);
```
arturdryomov, line 17: I don’t like the `apply_exclude_from` function as a side effect here.

src/remote_command.rs
```diff
@@ -0,0 +1,30 @@
+use config::Config;
+use std::process::Command;
+use std::process::Stdio;
+
+pub fn execute_remote_command(remote_command: &str, config: &Config, project_dir_on_remote_machine: &str) -> Result<(), ()> {
+    let mut command = Command::new("ssh");
+
+    command
+        .arg(config.remote_machine_name.clone())
```
arturdryomov, line 9: Why would you need `clone` here?

src/main.rs
```diff
@@ -0,0 +1,134 @@
...
+        }
+    }
+}
+
+fn execute_remote_command(working_dir: &PathBuf, args: &Args, config: &Config) -> Result<(), ()> {
+    println!("Executing command on remote machine...\n");
+
+    let start = Instant::now();
+
+    let result = execute_remote_command_impl(
```
arturdryomov: `impl`?

src/main.rs
```diff
@@ -0,0 +1,134 @@
...
+        &args.command.clone(),
+        config,
+        &format!("~/mainframer/{}", working_dir.file_name().unwrap().to_string_lossy().clone()),
```
arturdryomov: Don’t really like that the location is being repeated here and in the copy function.

arturdryomov (changes requested):

src/time.rs
pepyakin: FWIW, `assert_eq!` takes a `str` as the second argument. So you can use just `"0 seconds"`.

src/args.rs
artem-zinnatullin: nice! didn't know about this option, thx!

src/args.rs
artem-zinnatullin: oh I see the pattern here

artem-zinnatullin: I still like the `String::from` though, what do you think is more idiomatic?

String docs use `from()` heavily, that's why I used it https://doc.rust-lang.org/book/second-edition/ch08-02-strings.html

But `into()` seems to be a common function across different types as well, maybe I should use `from()` for "static" values and `into()` for dynamic?

src/args.rs
artem-zinnatullin: Heh, I would thought that Compiler will be able to optimize it here

src/args.rs
artem-zinnatullin: holy moly, this is now legit oneliner, THX!

src/main.rs
artem-zinnatullin: Rust doesn't support function overload :( :( :(

So I can't give these functions a same name even though arguments are completely different

src/main.rs
artem-zinnatullin: Yeah, good point, lemme refactor that, thx

src/main.rs
artem-zinnatullin: Do you have a better name for impl function on mind?

src/main.rs
```diff
@@ -0,0 +1,134 @@
...
+use remote_command::execute_remote_command as execute_remote_command_impl;
+use std::env;
+use std::path::PathBuf;
+use std::process;
+use std::time::Instant;
+use sync::*;
+use time::*;
+
+fn main() {
+    println!(":: Mainframer v2.1.0\n");
```
vanniktech: The version is duplicated here is there some way to pass in values into rust from tooling. e.g. custom defines in c++ or BuildConfig from Java.

src/remote_command.rs
artem-zinnatullin: Because I don't own `config` here, I only have a reference to it so I need to explicitly copy

https://doc.rust-lang.org/book/second-edition/ch04-01-what-is-ownership.html

src/main.rs
pepyakin: Yeah, cargo publishes package version thru build-time environment variables. 
https://doc.rust-lang.org/cargo/reference/environment-variables.html

src/sync.rs
artem-zinnatullin: result part is for sure pretty much same, will refactor, 
however I wouldn't combine sync logic as they use different `ignore` files and specify few `rsync` parameters completely different

src/sync.rs
artem-zinnatullin: Totally understand your frustration here as I'm also used to defensive copy everywhere especially with Kotlin

However, this is kinda whole point of Rust, it makes mutability compile-time safe. As you can see I pass `command` as `&mut` which clearly indicates that it can be mutated inside `apply_exclude_from()` function

Compiler checks that at any given point of time there is no data races and that only one piece of code owns the data, so such mutability is specifically built into the language to make it as fast as possible and avoid copying

Moreover, `Command` type is built to be mutated, method `arg()` actually modifies `Command` and returns `self` and not a new instance of `Command`

Same applies to all complex data types in Rust like `String` is also mutable, exception is only primitive types like `u32` that are typically cheaper to copy than mutate

src/sync.rs
artem-zinnatullin: wuuuuut, noice

src/time.rs
artem-zinnatullin: yep, will move

src/time.rs
artem-zinnatullin: doooope, thx!

src/time.rs
artem-zinnatullin: looks like so, that's how documentation suggests it and how other projects like ripgrep do it 

- https://doc.rust-lang.org/book/second-edition/ch11-01-writing-tests.html
- https://github.com/BurntSushi/ripgrep/blob/f5411b992cb5287e35ceca724651121c0beb3995/src/config.rs#L137

I've tried to move tests into `test` folder, but looks like it's suitable only for declaring common test code but not tests themselves

@pepyakin am I right?

src/time.rs
artem-zinnatullin: indeed!

test/test.sh
artem-zinnatullin: Yep for sure, let's do it in separate PR? I'd like to fail on warnings (currently 0) and enforce more checks like Clippy

src/main.rs
artem-zinnatullin: Yep yep, in plans in a separate PR, thanks for link!

src/time.rs
artem-zinnatullin: Can we please discuss that more? :) I think I'm missing something

Docs: https://doc.rust-lang.org/std/marker/trait.Copy.html

So as I see it - I can avoid copying if I don't need it because compiler allows me to do it safely, here with `duration: &Duration` I'm passing value by reference which only allocates additional reference that points to same struct

If I change function to `duration: Duration`, whenever I'd call it - a full copy of `Duration` struct will be allocated as well as new reference to it right?

If so, why should I copy if I don't really need a copy here?

pepyakin: Well, yeah, you're right. 

Maybe it's just me, my rule of thumb: if I have a `Copy` type then I just pass it by value since it in many cases is more convenient.

> If I change function to duration: Duration, whenever I'd call it - a full copy of Duration struct will be allocated as well as new reference to it right?

Actually it's very compilcated thing : ) But in the end, I think that there will be no difference in the generated machine code, since LLVM very likely will promote this reference to a value and pass it by a copy.

pepyakin: (As we say in russian: "Я парень простой, вижу `Copy` тип — передаю его по значению")

src/time.rs
pepyakin: Yeah, it is very common way to structure test code. And at times it is the only possible one, since submodules can access it's parent module items.

src/args.rs
pepyakin: I'm not aware of any convention here. `From` and `Into` are duals to each other in some sense.
You are free to use either `String::from(x)`, `x.to_string()`, `x.to_owned()` and `x.into()` and they will do the same thing, but I have an impression that `x.into()` is more common.

src/time.rs
artem-zinnatullin: hahah

Soo, indeed LLVM seems to be able to elide copy, however it's not guaranteed and there are cases where it doesn't do it

If you don't mind I'd keep it as `&Duration` for now, I'll need more Rust experience to get a full sense of when I want implicit copy or not :)

That was super helpful discussion though, really appreciate it :)

Useful links for myself:

- https://stackoverflow.com/questions/38571270/can-rust-optimise-away-the-bit-wise-copy-during-move-of-an-object-someday
- https://www.reddit.com/r/rust/comments/2xxjda/when_should_my_type_be_copy/

src/main.rs
artem-zinnatullin: @pepyakin how do you typically deal with this problem? I tend to prefer functionalish way of programming for code like this by avoiding unnecessary struct/objects and working with just top level functions

However in many cases that means that I have a "shell" version of a function that does some side-effects like printing to stdout, and sort of "internal" version of a function that is more pure

But Rust doesn't let my functions have same name :(

src/args.rs
artem-zinnatullin: Damn Rust, String is most confusing type so far, so many options jeez (and I haven't really used `Cow` yet btw)

Let's stick to `from()` for now since docs use it? I'd probably get a sense of convention as I read/write more code, googling didn't help much, I might also ask this later on /r/rust sub to get a sense of what community thinks about it :)

artem-zinnatullin:
@pepyakin btw feel free to "approve" or "request changes", that works even if you're not a member of the repo, your feedback is super valuable and I definitely owe you couple beers or something :D

src/time.rs
pepyakin: Heh, I meant the other way around: I bet that LLVM should convert references to a copies, because this usually opens new opportunities for optimizations (IIRC, this is called argument promotion). 

E.g in this code LLVM will promote the reference to the values, two values to be exact: `secs` and `nanos`. And since you don't use nanos in this code it will be elided altogether.

But yeah, LLVM also can elide big copies.

src/main.rs
pepyakin: It is broad problem I think, and sometimes to overcome rust lack of overloading it just comes down to choosing appropriate names.

However, in this case I can propose to use something like this

```rust
use remote_command;

remote_command::execute_remote_command(
  // ...
);
```

pepyakin: It's even not uncommon to choose name for inner function assuming it will be usually called thru the module.

One example that comes to my mind is `std::env::args()`. 
`args` is pretty ambigious name in itself. However, it is very common to call it like 

```rust
use std::env;

env::args()
```

src/time.rs
artem-zinnatullin: wuuut… it indeed can https://github.com/llvm-mirror/llvm/blob/master/lib/Transforms/IPO/ArgumentPromotion.cpp

okay, lemme switch to copy then lol

src/main.rs
artem-zinnatullin: neat! I like this way much better, will do same for sync, thanks 👍

Cargo.toml
arturdryomov: Do you want to open an issue about it?

ci/build.sh
```diff
@@ -20,9 +20,6 @@ if [ "$USER_ID" == "0" ]; then
     echo "Warning: running as r00t."
 fi
 
-# Run shellcheck.
-docker run --rm --env SHELLCHECK_OPTS="--exclude SC2088" --volume `"pwd"`:/scripts:ro koalaman/shellcheck:v0.4.6 /scripts/mainframer
```
arturdryomov, line 24: Thank God.

ci/docker/Dockerfile
arturdryomov: `chmod a+rwx`, please.

src/time.rs
```diff
@@ -0,0 +1,137 @@
...
+    fn format_duration_1_minute_1_second() {
+        assert_eq!(format_duration(&Duration::from_secs(61)), "1 minute 1 second");    }
+
+    #[test]
+    fn format_duration_1_minute_2_seconds() {
+        assert_eq!(format_duration(&Duration::from_secs(62)), "1 minute 2 seconds");    }
+
+    #[test]
+    fn format_duration_2_minutes() {
+        assert_eq!(format_duration(&Duration::from_secs(120)), "2 minutes 0 seconds");    }
```
arturdryomov: Formatting went off.

src/time.rs
```diff
@@ -0,0 +1,137 @@
...
+        }
+
+        push_value_and_label(&mut text, minutes, minutes_label);
+    }
+
+    if !text.is_empty() {
+        text.push(' ');
+    }
+
+    push_value_and_label(&mut text, seconds, seconds_label);
```
arturdryomov, line 44: You are still joining strings yourself...

arturdryomov (approved):

Cargo.toml
artem-zinnatullin: Yeeep, here you go #192

ci/docker/Dockerfile
artem-zinnatullin: 777, ruly?

@nostra13 sing it to @ming13:

Chmod777:

>Podnyal chmoda, stali drugimi prava
Stali schitatsya so mnoy
Znayut kto teper chown

>Podnyal chmoda, filesystem v liniu dala
U tebya talant bratan
Kkakoy?
Nastraivat linuxa

https://www.youtube.com/watch?v=qP303vxTLS8

src/time.rs
artem-zinnatullin: ?

src/time.rs
artem-zinnatullin: indeed

test/test.sh
artem-zinnatullin: Created #193 to track that

ci/build.sh
artem-zinnatullin: Well, actually I'm planning to add it back later to check our integration tests

artem-zinnatullin: Created #194 to track that

artem-zinnatullin:
Addressed remaining feedback, created few issues for changes I plan to do in separate PRs, going to merge it after CI

artem-zinnatullin:
@pepyakin SUPER BIG THANKS to you for review, that was very helpful and fun :kissing_cat: 

Gotta :beers: when we have chance!

@ming13 thanks for review as well :kissing_cat:

ci/docker/Dockerfile
```diff
@@ -22,14 +22,18 @@ ENV ANDROID_HOME /opt/android-sdk-linux
...
-    eval $ANDROID_SDK_INSTALL_COMPONENT '"tools"' && \
-    eval $ANDROID_SDK_INSTALL_COMPONENT '"platform-tools"' && \
-    eval $ANDROID_SDK_INSTALL_COMPONENT '"build-tools;25.0.2"' && \
-    eval $ANDROID_SDK_INSTALL_COMPONENT '"platforms;android-25"'
+    echo "Installing Android SDK components" && \
+    eval $ANDROID_SDK_INSTALL_COMPONENT '"tools"' > /dev/null && \
+    eval $ANDROID_SDK_INSTALL_COMPONENT '"platform-tools"' > /dev/null && \
+    eval $ANDROID_SDK_INSTALL_COMPONENT '"build-tools;25.0.2"' > /dev/null && \
+    eval $ANDROID_SDK_INSTALL_COMPONENT '"platforms;android-25"' > /dev/null && \
+    chmod -R a+rwx "$ANDROID_HOME"
```
nostra13, line 36: > Vse govoryat "AZ, a kak podnyat prava?"
> "Kuda nazhat'?" - "Ch-Ch-cmod tri topora"
> Sudo ne pishu, paroli ne vvozhu.
> Vse, chto podnyal vchera -
> Paru instansov dokera.

artem-zinnatullin: suk ubil

artem-zinnatullin: :heart: u
miss u

Issue #192, opened 2018-03-01
Resolve actual version from git tag during build.


artem-zinnatullin:
Currently version is stored in project config which require us to update it manually and commit every time, we should get it from current git tag during build as we do for other projects like:

- https://github.com/gojuno/composer
- https://github.com/gojuno/koptional

and others which allows us release new versions just by pushing new tag or through GitHub releases.

artem-zinnatullin:
Related: https://github.com/gojuno/mainframer/pull/191/files#r170609701

artem-zinnatullin:
Closed by #203.

PR #195, opened 2018-03-01
Resolve Mainframer version during build.

artem-zinnatullin:
Part of #192.

Docs: https://doc.rust-lang.org/cargo/reference/environment-variables.html

Thanks @pepyakin for the link!

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -14,7 +14,7 @@ use std::time::Instant;
 use time::*;
 
 fn main() {
-    println!(":: Mainframer v2.1.0\n");
+    println!(":: Mainframer v{}\n", env!("CARGO_PKG_VERSION"));
 
     let args = match Args::parse(env::args().skip(1).collect()) {
         Err(message) => exit_with_error(&message, 1),


vanniktech (approved):
Nice

artem-zinnatullin:
@vanniktech appreciate your interest and reviews 😽

src/main.rs
```diff
@@ -14,7 +14,7 @@ use std::time::Instant;
 use time::*;
 
 fn main() {
-    println!(":: Mainframer v2.1.0\n");
+    println!(":: Mainframer v{}\n", env!("CARGO_PKG_VERSION"));
```
arturdryomov, line 17: How does it even work though? When runtime is packaged as a binary it exports environment variables itself?

artem-zinnatullin: `env!` is a macro that compiler executes at build time, `CARGO_PKG_VERSION` is an environment variable provided by `cargo` to `rustc` during the build, so as a result we get version inlined into the code and binary

arturdryomov: These native folks will steal your soul eventually, you know. Avoid blood-involved rituals at all costs.

arturdryomov (approved):

src/main.rs
artem-zinnatullin: lol, will try

btw Kotlin might get macros support eventually

Issue #193, opened 2018-03-01
Make Rust build stricter.


artem-zinnatullin:
Specifically:

- Fail build if compiler gives any warnings
- Look into integrating https://github.com/rust-lang-nursery/rust-clippy

artem-zinnatullin:
Hmm, looks like failing build because of warnings is not considered a super great practice in Rust community (cc @pepyakin):

- https://github.com/rust-unofficial/patterns/blob/master/anti_patterns/deny-warnings.md
- https://users.rust-lang.org/t/how-to-fail-a-build-with-warnings/2687/9

As far as I understand it's mainly due to the way you get Rust on your system, by default `rustup` brings whatever latest stable version is which might bring new warnings and break build that didn't change any code

However there are ways to install specific Rust version, so combining that with failing on warnings seems reasonable for me because I want to control Rust version anyway

PR #196, opened 2018-03-01
Use fixed version of Rust (1.24.0) instead of latest stable.

artem-zinnatullin:
Part of #193.

This makes our build much more predictable and reproducible.


arturdryomov:
Nothing says safety better than running bash scripts from a web link.

arturdryomov (approved):

artem-zinnatullin:
Hey, at least it's https lol, but that's recommended install way https://www.rust-lang.org/en-US/install.html

artem-zinnatullin:
Btw, highly recommend to check code of https://static.rust-lang.org/rustup.sh

That's level of programming in Bash I can live with

arturdryomov:
The script looks nice, but such size makes it kind of complex anyway ;-(

artem-zinnatullin:
Well sure, but you probably don't want script to download its parts during execution

PR #216, opened 2018-03-29
Setup deployment to GitHub releases.

### LINKED ISSUE #198: Deploy Mainframer binaries to GitHub releases.
Since we now use compiled language #191, we need to ship binaries instead of source code.

To do that I'm planning to deploy resulting binaries to GitHub releases which can be either manually downloaded or installed through Homebrew/etc. Later we could also upload binaries to Bintray to have them there just in case.

As a reference example, this is how `ripgrep` does it:

- [ripgrep GitHub releases page](https://github.com/BurntSushi/ripgrep/releases)
- [ripgrep Homebrew](https://github.com/BurntSushi/ripgrep/blob/f5411b992cb5287e35ceca724651121c0beb3995/pkg/brew/ripgrep-bin.rb)
- [ripgrep Travis config that deploys to GitHub](https://github.com/BurntSushi/ripgrep/blob/f5411b992cb5287e35ceca724651121c0beb3995/.travis.yml#L89)

And Travis documentation about deploying to GitHub releases https://docs.travis-ci.com/user/deployment/releases/

artem-zinnatullin:
Closes #198.

I don't really like it yet, will probably refactor later.

So far to distinguish macOS/Linux binaries I'm just adding `uname -s` to the binary name. 

But if we'll need to build more than one targets for same OS, we'll need to add target info to the file name.

I'm also not sure if I want to add Mainframer version to the binary name:

- It's not easy, I'll need to do it from Bash, `cargo` doesn't do that
- It's better for raw downloads because you see version in the file name
- It's worse for raw downloads because you can't just put it in `$PATH` (but I already broke it by adding `uname -s`)
- It's worse for Homebrew tap as we'll need to reference different file name in each release


artem-zinnatullin:
Gonna test deployment on my fork

artem-zinnatullin:
```
> Could not resolve all files for configuration ':classpath'.
   > Could not download gradle-core.jar (com.android.tools.build:gradle-core:2.2.3)
      > Could not get resource 'https://jcenter.bintray.com/com/android/tools/build/gradle-core/2.2.3/gradle-core-2.2.3.jar'.
         > Could not GET 'https://jcenter.bintray.com/com/android/tools/build/gradle-core/2.2.3/gradle-core-2.2.3.jar'.
            > Connect to akamai.bintray.com:443 [akamai.bintray.com/23.45.130.89] failed: Connection timed out (Connection timed out)
```

Of course Bintray [reports that everything is ok](http://status.bintray.com).

screen shot 2018-03-29 at 9 42 55 am

.travis.yml
```diff
@@ -3,16 +3,26 @@ install:
...
-        - ci/macOS.sh
+  - os: linux
+    sudo: required
+    services:
+    - docker
+    script:
+    - ci/build.sh
+  - os: osx
+    script:
+    - ci/macOS.sh
```
arturdryomov, line 14: `build-macos.sh`?

arturdryomov (approved):
We should move samples to a separate repository with multiple Docker images since the current setup is a mess.

artem-zinnatullin:
Once we publish Mainframer to at least Homebrew, I'm not against moving samples to a separate repo

.travis.yml
artem-zinnatullin: I want to refactor it separately, it's a little messy atm :(

artem-zinnatullin:
Works on my fork: https://github.com/artem-zinnatullin/mainframer/releases/tag/v3.0.0-deploy-test-9

PR #229, opened 2018-10-15
Reset Apache license to the original state.

arturdryomov:


artem-zinnatullin:
Motivation: last section in the license text is a template, not an actual part that we need to change

artem-zinnatullin (approved):

artem-zinnatullin:
mhm, Travis integration didn't migrate after project transfer, fixing…

Issue #207, opened 2018-03-20
Update documentation for 3.x.


artem-zinnatullin:
Most of documentation is actually compatible with 3.x, but specifically section about downloading and keeping Mainframer will need to be updated as we switch to distribution through Homebrew and other package managers #32 and OS-wide usage #185.

arturdryomov:
I have a radical proposition to streamline things and remove the baggage.

The proposed documentation structure:

```
.
└── docs
    ├── configuration
    │   ├── mainframe.md
    │   ├── user-examples.md
    │   └── user.md
    ├── integrations
    │   └── intellij-idea.md
    └── setup
        ├── mainframe.md
        └── user.md
```
* As you can see I’ve replaced _local_ and _remote_ terms with _user_ (as user terminal) and _mainframe_. Over time I feel that everybody invents their own name for machines and we might as well bring our own terms more in line with the Mainframer naming. Hopefully it will stick better.
* Recipes are gone since the naming was wrong all along. In my understanding recipes are more about _how to use the tool_ instead of _how to use other tools with the tool_.
* Samples are gone for a couple of reasons: it is painful to support them, as tests they play zero to none purpose (otherwise the test suite is bad) and, more importantly, GitHub traffic stats show that people do not open them (except the Android one but we have Android Studio docs here for some reason). I’ve replaced them with`configuration/user-examples.md` which will contain various `ignore.yml` which actually play a purpose in samples.
* Not to worry — `configuration/mainframe.md` will contain a single phrase like _Mainframe does not have any configuration except SSH`.

@artem-zinnatullin, if you agree, I can start implementing this.

artem-zinnatullin:
In general I'm in favor of change, but I do have comments

- Definitely not a fan of "mainframe" and "user" (wut), I'd like us to be consistent with configs and use same terminology, like "remote" and "local"
- "user-examples" should be a directory similar to "integrations", imo, it looks weird in "configuration" dir

arturdryomov:
Thought about configs as well but we use `push` and `pull` already.

artem-zinnatullin:
Damn it

artem-zinnatullin:
Wait, we still have "remote" in config, let's keep "remote" and "local" then?

PR #230, opened 2018-10-15
Fix samples/mvn/README.md after #185

szymonm:
After introducing #185 mainframer is a system wide command -- updating readme of mvn sample.

### DIFF
--- a/samples/mvn/README.md
+++ b/samples/mvn/README.md
@@ -7,7 +7,7 @@ This is a very common setup for basically any Maven project (including multimodu
 ### How to build
 
 ```bash
-$ ./mainframer ./mvnw package
+$ mainframer ./mvnw package
 ```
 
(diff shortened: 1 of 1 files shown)


artem-zinnatullin (approved):

szymonm:
Will you merge it @artem-zinnatullin ?

artem-zinnatullin:
Waiting for @ming13 review :)

We have same `./mainframer` instead of `mainframer` going on in the other samples. 

@ming13 I think it's good time for us to start migrating basic docs to 3.x (see #207), we can point people on 2.x tags or 2.x branch if they need docs for 2.x, most of the docs are backwards/forward compatible anyway

artem-zinnatullin:
Alright, gonna merge this and update the remaining examples

This particular change is not really breaking for the documentation as it just shows how to run something with Mainframer and could have been true with 2.x if user has an alias

PR #236, opened 2018-11-14
Replicate local absolute path on remote machine

### LINKED ISSUE #214: Replicate path to the project from local machine on remote machine.
At the moment Mainframer 2.x creates following path on remote machine:

Example:

```console
local machine path : ~/projects/mycompany.com/project-x
remote machine path: ~/mainframer/project-x
```

As we scale Mainframer for OS-wide usage #185, chances of folder name clashing increase.

Example:

```console
local machine path 1: ~/projects/mycompany.com/project-x
local machine path 2: ~/project-x

remote machine path 1: ~/mainframer/project-x
remote machine path 2: ~/mainframer/project-x
```

In the example above projects with different path on local machine end up in same path on remote machine.

This is unfortunate and unexpected by user.

I propose following simple solution: include part of local machine project path after `$HOME` into remote path.

Example:

```console
local machine path 1: ~/projects/mycompany.com/project-x
local machine path 2: ~/project-x

remote machine path 1: ~/mainframer/mycompany.com/project-x
remote machine path 2: ~/mainframer/project-x
```

---

**Important note**: this change is incompatible with Mainframer 2.x.
However it'll just work in many cases, but users will have old unused folder on remote machine.

@ming13 wdyt?

artem-zinnatullin:
Closes #214.

As discussed in #214, we now replicate absolute path from local machine in `~/mainframer/…` on remote machine.

This allows user to have multiple projects with same leaf node in the path without clashing, ie:

- `/home/artem_zin/myproject` → `/home/remote_user/mainframer/home/artem_zin/myproject`
- `/home/artem_zin/projects/myproject` → `/home/remote_user/mainframer/home/artem_zin/projects/myproject`

We can also add local machine hostname, but I don't think it's a good idea.

---

Integration tests were updated to expect files in new path. Also tested myself.


src/sync.rs
```diff
@@ -46,15 +47,15 @@ pub fn sync_remote_to_local(local_project_dir_name: &str, config: &Config, ignor
...
         )
         .arg("./");
 
     execute_rsync(&mut command)
 }
 
-fn project_dir_on_remote_machine(local_project_dir_name: &str) -> String {
-    format!("~/mainframer/{}", local_project_dir_name)
+pub fn project_dir_on_remote_machine(local_dir_absolute_path: &Path) -> String {
+    format!("~/mainframer/{}", local_dir_absolute_path.to_string_lossy())
```
dmitry-novikov: Could it be here `"~/mainframer//home/user/project"` as a result?

artem-zinnatullin: Snap, good catch 👍 

Most of the time it's harmless to do so (you can check in your terminal), but I definitely did not intend that

artem-zinnatullin: foxed, thanks again 😽

arturdryomov (approved):

Issue #213, opened 2018-03-25
User-level config files in $HOME directory.


artem-zinnatullin:
As we're making Mainframer OS-wide tool #185, we need to add support for configuration files stored in user-home `$HOME` directory that will be used in case when work-dir doesn't have them.

Purpose of this feature is to enable use cases like:

```console
$ git clone newproject
$ cd newproject
$ mainframer make build
```

When you have project(s) you don't want to configure specifically and just let Mainframer do its work with 0-setup on project level.

Current plan:

- Use `~/.mainframer` folder as user-home level folder for config files (`config`, `ignore`, `localignore`, `remoteignore`)
- Currently we don't plan to merge user-level and project-level configs, project-level will always be prioritized and used instead

There is an existing PR #189 that was targeting 2.x version of Mainframer and not fully compatible with 3.x rewritten in Rust, so we'll merge parts of #189 (specifically integration tests) and add another PR to implement this functionality in Mainframer.

@ming13 do you agree with details like `~/.mainframer` as folder name and non-merge behavior?

arturdryomov:
Please don’t forget about `XDG_CONFIG` (Linux-specific).

On other note, I think we should actually merge configs like Git does. Take global first, merge with local and use the result. It is not that hard actually, especially since we have a _real_ programming language to use.

Pozzoooo:
@ming13 The integration tests are actually testing properly merged config file like you described, though for the ignore files it is using the local and fall backing for the shared ones, so no merge tests for ignores.

Also, there is a small challenge on this approach, as user could run mainframer at any folder, it might happen to run in the wrong folder.
It happened to me some times, I end up syncing stuff I didn't wanted to, and just realize it after the command fail as there was no "gradlew" or something like this (it happend some times on home folder, just terrible :laughing: ).
I've tested a few ideas on my fork, the one I liked the most so far is prompt the user to confirm when it is happen to be a new folder.

artem-zinnatullin:
Merging is not that easy tho, specifically because we don't have a format to unignore stuff like git does:

```
.idea
!.idea/codeStyleSettings.xml
```

PR #239, opened 2018-11-26
Switch config to YAML

### LINKED ISSUE #225: Configuration files format.
Just want to get discussion around configuration files format we use.

Right now it's https://en.wikipedia.org/wiki/.properties, which is fast to parse and easy to read & write.

However it's not super common among other tools and it doesn't have a standartized way to have sections in it and it doesn't look very structured. However I like that it maps 1:1 with command-line args we'll be adding in 3.x.

Possible alternatives:

- [YAML](https://en.wikipedia.org/wiki/YAML) (not a fan, but it's very popular nowadays)
- [JSON](https://en.wikipedia.org/wiki/JSON) (doesn't have comments but easy to work with)
- [INI](https://en.wikipedia.org/wiki/INI_file) (very similar to `.properties`, but has sections, also not very popular)

Or we can keep everything as is of course :)

Wonder what our users think, especially given that in 3.x we'll introduce global config files and maybe couple more. cc @ming13

artem-zinnatullin:
Closes #225, part of #213 

This change switches config format that used to be some sort of `.ini` file to [YAML](http://yaml.org/spec/1.2/spec.html#id2760395) as discussed in #225 

Example of `.mainframer/config.yml`:

```yml
remote:
  host: "computer1"
  user: "user1" # Not implemented yet.
  port: 1234    # Not implemented yet.
push:
  compression: 5
pull:
  compression: 2
```

---

Code is also getting ready for config files merging #213. 

- `IntermediateConfig` represents one of possibly multiple config files, thus everything is optional.
- `Config` represents finalized config, thus only actually optional fields are optional

Note that `remote.user` and `remote.port` are not implemented yet (we continue to rely on `~/.ssh/config` read by `ssh` for now)

---

Description was updated to reflect changes after review, you can see edit history.


src/intermediate_config.rs
```diff
@@ -0,0 +1,305 @@
...
+    pub compression: Option<IntermediateCompression>,
+}
+
+#[derive(Debug, Eq, PartialEq)]
+pub struct IntermediateRemoteMachine {
+    pub host: Option<String>,
+}
+
+#[derive(Debug, Eq, PartialEq)]
+pub struct IntermediateCompression {
```
arturdryomov: Shouldn’t we use the same value for both channels? I’m already forgetting what is the benefit of having different ones. Yes, it can be a bit faster to send sources uncompressed, but I believe compression does not eat a lot of resources in the first place.

src/config.rs
```diff
@@ -1,161 +1,16 @@
...
-
-    #[test]
-    fn parse_config_from_str_remote_compression_level_not_a_number() {
-        let content = "remote_machine=test@machine\nremote_compression_level=wut";
-        assert_eq!(parse_config_from_str(content), Err(String::from("'remote_compression_level' must be a positive number, found 'wut'.")));
-    }
+#[derive(Debug, Eq, PartialEq)]
+pub struct Compression {
+    pub local: i64,
+    pub remote: i64,
```
arturdryomov: Shouldn’t it be `i8` though?

src/intermediate_config.rs
```diff
@@ -0,0 +1,305 @@
...
+                host,
+            })
+        }
+        Yaml::Null | Yaml::BadValue => None,
+        ref something_else => return Err(format!("'remoteMachine' must be an object, but was {:#?}", something_else))
+    };
+
+    let compression = match &yaml["compression"] {
+        Yaml::Hash(compression) => {
+            let local = match compression.get(&Yaml::String(String::from("local"))).cloned() {
```
arturdryomov: Dude, YAML parsing in Rust sucks!

src/intermediate_config.rs
```diff
@@ -0,0 +1,305 @@
...
+    use super::*;
+
+    #[test]
+    fn parse_config_from_str_all_fields_2_spaces_indent() {
+        let content = "
+remoteMachine:
+  host: computer1
+compression:
+  local: 5
+  remote: 2"
```
arturdryomov: All right, in the PR description you mention `user` and `port`, but both seem non-existant in the implementation. And I guess we don’t need them since SSH handles it for us.

I think we have not discussed config contents yet. I see it this way.

```yaml
# remote, like Git
remote:
  name: "SSH name"
sending:
  compression: N
  ignore:
    - "path"
    - "path"
receiving:
  compression: N
  ignore:
    - "path"
    - "path"
# compression declared this way overrides sending and receiving values
compression: N
# compression declared this way extends sending and receiving values
ignore:
  - "path"
  - "path"
```

Do not forget that `.mainframer.yml` can be placed both at `${HOME}` and at `${PROJECT}`.

* Corporate mainframe, multiple projects.
    * Home config — basically everything. Project config — additional ignore rules.
* Corporate and personal mainframe, multiple projects.
    * Home config — personal thingies. Project config — additional ignore rules plus different remote and probably compression levels.

Basically what Git does, but with sharing in VCS.

And yes, I think we should move away from a `.mainframer` directory since a single file simplifies things.

arturdryomov (changes requested):

src/intermediate_config.rs
artem-zinnatullin: Text is very compressible compared to compiled binaries, thus we recommend at least `3` for local compression and `1` for remote

But obviously, each project can be different — thus it's a setting

As well as different use cases, for example when you have limited connection you might prefer compression over latency

It does eat quite a lot of resources, try `9` :)

src/config.rs
artem-zinnatullin: yea, it's a slip of YAML impl details, I guess I should use proper type here

src/intermediate_config.rs
artem-zinnatullin: well, could have been worse lol

src/intermediate_config.rs
artem-zinnatullin: From the description:

```
Note that `remoteMachine.user` and `remoteMachine.port` are not implemented yet (we continue to rely on `~/.ssh/config` read by `ssh` for now)
```

Also, I've filed #240 and #241 for user and port properties (there is description explaining why)

artem-zinnatullin: >And yes, I think we should move away from a .mainframer directory since a single file simplifies things.

That we can't do easily, similarly to `.gitignore` and `.git/config` you want to check in ignore lists into version control, while not checking in the config

artem-zinnatullin: >Do not forget that .mainframer.yml can be placed both at ${HOME} and at ${PROJECT}.

I know lol, that's why I want to merge it so badly to implement $HOME part

arturdryomov:
After internal discussion we decided to pursue the following.

### `.mainframer/config.yml`

```yaml
remote:
  name: "name"
push:
  compression: N
  sync: "serial"
pull:
  compression: N
  sync: "parallel"
```

### `.mainframer/ignore.yml`

```yaml
both:
  - "path"
  - "path"
  - "path"
push:
  - "path"
  - "path"
  - "path"
pull:
  - "path"
  - "path"
  - "path"
```

src/intermediate_config.rs
artem-zinnatullin: Done

src/config.rs
artem-zinnatullin: Not done yet, I'll track it separately, value is validated anyway and realistically i8 is going to still be i32 at least on actual hardware

artem-zinnatullin: #243

artem-zinnatullin:
@ming13 updated PR with new YAML structure

src/config.rs
```diff
@@ -1,16 +1,21 @@
...
+pub struct Remote {
     pub host: String,
 }
 
 #[derive(Debug, Eq, PartialEq)]
-pub struct Compression {
-    pub local: i64,
-    pub remote: i64,
+pub struct Push {
+    pub compression: i64,
```
arturdryomov: Still — `i64` is too much.

src/config.rs
```diff
@@ -1,16 +1,21 @@
...
+}
+
+#[derive(Debug, Eq, PartialEq)]
```
arturdryomov: Why do you need both `Eq` and `PartialEq`?

src/intermediate_config.rs
```diff
@@ -8,19 +8,24 @@ use self::yaml_rust::YamlLoader;
 
 #[derive(Debug, PartialEq, Eq)]
 pub struct IntermediateConfig {
-    pub remote_machine: Option<IntermediateRemoteMachine>,
-    pub compression: Option<IntermediateCompression>,
+    pub remote: Option<IntermediateRemote>,
+    pub push: Option<IntermediatePush>,
+    pub pull: Option<IntermediatePull>,
```
arturdryomov, line 15: Really weird, but I like that both `push` and `pull` have the same horizontal space...

src/intermediate_config.rs
```diff
@@ -48,65 +53,76 @@ fn parse_config_from_str(config_content: &str) -> Result<IntermediateConfig, Str
...
+        ref something_else => return Err(format!("'remote' must be an object, but was {:#?}", something_else))
     };
 
-    let compression = match &yaml["compression"] {
-        Yaml::Hash(compression) => {
-            let local = match compression.get(&Yaml::String(String::from("local"))).cloned() {
-                Some(local) => match local {
-                    Yaml::Integer(local) => if local >= 1 && local <= 9 {
-                        Some(local)
+    let push = match &yaml["push"] {
```
arturdryomov, line 77: I think it is time to move `push` and `pull` reading to a single function since essentially they are the same. Maybe even replace `struct Pull` and `struct Push` with `struct Sync` or whatever.

src/intermediate_config.rs
```diff
@@ -8,19 +8,24 @@ use self::yaml_rust::YamlLoader;
...
+pub struct IntermediateRemote {
     pub host: Option<String>,
 }
 
 #[derive(Debug, Eq, PartialEq)]
-pub struct IntermediateCompression {
-    pub local: Option<i64>,
-    pub remote: Option<i64>,
+pub struct IntermediatePush {
+    pub compression: Option<i64>
```
arturdryomov: `compressionLevel`?

arturdryomov (changes requested):

src/intermediate_config.rs
artem-zinnatullin: They're not same, `Pull` will support `parallel` mode soon, common fields can be extracted though

src/config.rs
artem-zinnatullin: Switched to `u8`

src/config.rs
artem-zinnatullin: That's a great question my friend!

I need equality for unit tests (`assert_eq!()`)
In Rust `Eq` seems to be special case of `PartialEq` and most of things like `assert_eq!()` need `PartialEq` thus I definitely need to  have `PartialEq`

Removed `Eq` 👍

src/intermediate_config.rs
artem-zinnatullin: Why though, it's called `compression` in YAML, I mean we can have different name in code but I don't see much value in that atm

artem-zinnatullin:
Updated PR, @ming13 let's do YAML for ignore files in separate PR, it will be a pretty big change

arturdryomov:
All right, I’ll try not to leave any comments since the next time you’ll have time to make changes is approximately March 26.

src/config.rs
```diff
@@ -1,161 +1,21 @@
-use std::fs::File;
-use std::io::prelude::*;
-use std::path::Path;
-
-#[derive(Debug, PartialEq, Eq)]
+#[derive(Debug, PartialEq)]
```
arturdryomov, line 1: I still have no idea why language has `PartialEq` *and* `Eq`.

Cargo.toml
```diff
@@ -9,3 +9,5 @@ version = "3.0.0-dev"
 authors = ["Artem Zinnatullin <artem.zinnatullin@gmail.com>", "Artur Dryomov <artur.dryomov@gmail.com>", "Mainframer Developers and Contributors"]
 
 [dependencies]
+yaml-rust = "0.4.2"
+linked-hash-map = "0.5.1"
```
arturdryomov, line 13: Wait, what? No maps in Rust?

src/intermediate_config.rs
```diff
@@ -0,0 +1,330 @@
...
+    pub fn from_file(file_path: &Path) -> Result<IntermediateConfig, String> {
+        let mut content = String::new();
+
+        let mut file = match File::open(file_path) {
+            Err(_) => return Err(format!("Could not open config file '{}'", file_path.to_string_lossy())),
+            Ok(value) => value,
+        };
+
+        file.read_to_string(&mut content)
+            .unwrap_or_else(|_| panic!("Could not read config file '{}'", file_path.to_string_lossy()));
```
arturdryomov, line 43: *Cannot*?

src/intermediate_config.rs
```diff
@@ -0,0 +1,330 @@
...
+                }),
+                Err(error) => return Err(error)
+            }
+        }
+        Yaml::Null | Yaml::BadValue => None,
+        _ => return Err(String::from("'push' must be an object"))
+    };
+
+    let pull = match &yaml["pull"] {
+        Yaml::Hash(pull) => {
```
arturdryomov, line 93: Again, YAML parsing is so weird...

src/intermediate_config.rs
```diff
@@ -0,0 +1,330 @@
...
+  host: computer1
+push:
+  compression: 5
+pull:
+  compression: 2
+";
+
+        assert_eq!(parse_config_from_str(content), Ok(IntermediateConfig {
+            remote: Some(IntermediateRemote {
+                host: Some(String::from("computer1")),
```
arturdryomov, line 146: Isn’t it better to define expected variables beforehand and put them into the `content`?

src/intermediate_config.rs
```diff
@@ -0,0 +1,330 @@
...
+    }
+
+    #[test]
+    fn parse_config_from_str_pull_compression_remote_not_an_integer() {
+        let content = "
+pull:
+  compression: yooo
+";
+        assert_eq!(parse_config_from_str(content), Err(String::from("'pull.compression\' must be a positive integer from 1 to 9, but was String(\n    \"yooo\"\n)")));
+    }
```
arturdryomov, line 329: Pretty good test suite.

arturdryomov (approved):

PR #245, opened 2018-12-29
Add reworked configuration docs.

arturdryomov:


docs/configuration/description.md
```diff
@@ -0,0 +1,67 @@
+Files are placed in the `.mainframer` directory.
+The final configuration is the combination of files placed in:
+
+* `${HOME}/.mainframer` — the global configuration;
```
artem-zinnatullin: btw, that I would probably like to change to `~/.config/mainframer` 

A lot of other tools store configs there (example from my machine):

```console
tree -L 1 ~/.config
/Users/user/.config
├── alacritty
├── gcloud
├── gtk-2.0
├── htop
├── kitty
├── nvim
└── thefuck
```

Can't find it being described as some standard other than by XDG https://unix.stackexchange.com/questions/68721/where-should-user-configuration-files-go but a lot of tools seem to use it now

docs/configuration/description.md
```diff
@@ -0,0 +1,67 @@
...
+* `.mainframer` — the project configuration.
+
+It is posssible to share the configuration using the global one
```
artem-zinnatullin: Maybe reword?

"It is possible to declare common configuration using the global one and fine-tune it using the project one"

docs/configuration/description.md
```diff
@@ -0,0 +1,67 @@
...
+  name: "{SSH machine name}"
+push:
+  compression: {level}
+pull:
+  compression: {level}
+```
+
+Name               | Optional | Value   | Default | Description
+-------------------|----------|---------|---------|------------------
+`remote.name`      | No       | `string`| —       | Remote machine name from SSH config.
```
artem-zinnatullin: "Hostname/IP or remote machine name from `$HOME/.ssh/config`."

docs/configuration/description.md
```diff
@@ -0,0 +1,67 @@
...
+```yaml
+remote:
+  name: "{SSH machine name}"
+push:
+  compression: {level}
+pull:
+  compression: {level}
+```
+
+Name               | Optional | Value   | Default | Description
```
artem-zinnatullin: Usually it's "Required" instead of "Optional" :)

docs/configuration/description.md
```diff
@@ -0,0 +1,67 @@
...
+`pull.compression` | Yes      | `0..9`  | `0`     | Compression level used to copy files from remote machine to local one.
+
+Compression level is inherited from underlying `rsync`
+which uses [`zlib` values](https://www.zlib.net/manual.html):
+
+> The compression level must be between `0` and `9`:
+> `1` gives best speed, `9` gives best compression, `0` gives no compression at all
+> (the input data is simply copied a block at a time).
+
+## `.mainframer/ignore.yml`
```
artem-zinnatullin, line 46: This section misses description

docs/configuration/description.md
arturdryomov: Sounds good. Yes, it is XDG-only, but I don’t mind a semi-clean home directory on macOS.

docs/configuration/description.md
arturdryomov: Files are described in the `.mainframer` section above.

docs/configuration/description.md
artem-zinnatullin: My example is from macOS just in case :)

docs/configuration/description.md
artem-zinnatullin: Yeah, I guess I just wanted to see description there, ok let's go without it for now then

arturdryomov:
@artem-zinnatullin, updated.

docs/configuration/description.md
```diff
@@ -0,0 +1,69 @@
+Files are placed in the `.mainframer` directory.
```
artem-zinnatullin: This sentence is overlapped with next paragraph and also is incorrect in case of `${HOME}/.config/mainframer`

docs/configuration/description.md
```diff
@@ -0,0 +1,69 @@
...
+push:
+  compression: {level}
+pull:
+  compression: {level}
+```
+
+Name               | Required | Value   | Default | Description
+-------------------|----------|---------|---------|------------------
+`remote.name`      | Yes      | `string`| —       | Remote machine name from SSH config or hostname / IP address.
+`push.compression` | No       | `0..9`  | `0`     | Compression level used to copy files from local machine to remote one.
```
artem-zinnatullin, line 36: default is 3

docs/configuration/description.md
```diff
@@ -0,0 +1,69 @@
...
+`pull.compression` | No       | `0..9`  | `0`     | Compression level used to copy files from remote machine to local one.
```
artem-zinnatullin, line 37: default is 1

artem-zinnatullin (approved):

docs/configuration/description.md
arturdryomov: Are these our values? Do you have a reference otherwise? I’ve found [an outdated email](http://samba.2283325.n4.nabble.com/what-is-the-default-value-for-compress-level-td2538905.html). [The source code says otherwise](https://github.com/kuna-systems/rsync/blob/5546dab32970955e77ef7a5886bcd8fb765a25bf/options.c#L1825-L1850) — seems to be `-1` (`Z_DEFAULT_COMPRESSION` value).

arturdryomov: MF `2.x` uses `1` by default for both cases.

artem-zinnatullin: It's in 3.x PR #239, idea is that usually you push text files which are very compressible and pull binaries which are barely compressible, thus push `3` and pull `1`

artem-zinnatullin: We can update this doc in #239, going to merge your PR now

PR #247, opened 2019-01-12
Add reworked getting started guide for remote machine.

arturdryomov:


docs/getting-started/remote-machine.md
```diff
@@ -0,0 +1,102 @@
...
+```bash
+# If not running interactively, don't do anything
+```
+
+We’ve built a Bash script which might help with creating Linux users.
+
+Usage:
+
+```console
+$ bash create-user.sh "USER_NAME" "USER_SSH_PUBLIC_KEY"
```
artem-zinnatullin, line 34: no need for `bash`, script has proper shebang line already, just `./create-user.sh …`

artem-zinnatullin (approved):

docs/getting-started/remote-machine.md
arturdryomov: Well, you have to make it executable beforehand, which will result in another command here.

PR #248, opened 2019-01-12
Add configuration examples docs.

arturdryomov:


docs/configuration/examples.md
```diff
@@ -0,0 +1,104 @@
+# Configuration Examples
+
+## `ignore.yml`
+
+### General
+
+```yaml
+push:
+  # No need to sync build results to the remote machine.
```
artem-zinnatullin: Can we reword this into something more neutral or something? 

"No need" reads weird because it kind of implies that this configuration is not needed or something lol

Maybe "Ignore build results from local machine when pushing to remote."?

docs/configuration/examples.md
```diff
@@ -0,0 +1,104 @@
...
+  # Replace "out" with the directory where build results are placed.
+  - "out"
+pull:
+  # No need to sync sources to the local machine.
```
artem-zinnatullin: Same: "Ignore sources from remote machine when pulling to local."

docs/configuration/examples.md
```diff
@@ -0,0 +1,104 @@
...
+  # Replace "src" with the directory where the source code is placed.
+  - "src"
+```
+
+### Tools
```
artem-zinnatullin, line 18: "Tool-specific configurations"?

docs/configuration/examples.md
```diff
@@ -0,0 +1,104 @@
...
+
+#### IntelliJ IDEA
+
+Includes Android Studio, PyCharm, WebStorm, GoLand and other tools built
+on the IntelliJ Platform.
+
+```yaml
+push:
+  # Synching IDE-specific directories is not required for the build.
+  - "/.idea"
```
artem-zinnatullin, line 36: You can also add `*.iml`, because they're all over the place in IntelliJ projects

docs/configuration/examples.md
```diff
@@ -0,0 +1,104 @@
...
+
+### Build Systems
+
+#### Buck
+
+```yaml
+push:
+  # No need to sync build results to the remote machine.
+  - "buck-out"
+both:
```
artem-zinnatullin, line 59: We'll need to find another word for this one lol

docs/configuration/examples.md
```diff
@@ -0,0 +1,104 @@
...
+```
+
+#### Gradle: Android
+
+Gradle rules apply as well.
+
+```yaml
+push:
+  # Syncing captures from Android Studio is usually not required.
+  - "/captures"
```
artem-zinnatullin: Why is it in Gradle section though, this is just Android Studio and it'll do same for Bazel projects afaik

artem-zinnatullin (approved):

docs/configuration/examples.md
arturdryomov: *Making a GLaDOS voice*
Now you are thinking with Bazel.

PR #249, opened 2019-01-12
Add IntelliJ IDEA integration docs.

arturdryomov:


docs/integration/intellij-idea.md
```diff
@@ -0,0 +1,56 @@
+# Integration with IntelliJ IDEA
+
+Both plugin and plugin-less integrations use Run Configurations
```
artem-zinnatullin, line 3: Put "Run Configurations" in quotes?

artem-zinnatullin (approved):

docs/integration/intellij-idea.md
arturdryomov: I’ve tried it but it kinda forces using quotes in all occurrences which pollutes the text visually.

PR #252, opened 2019-01-19
Replace CI Docker image with Travis environment.

arturdryomov:
* Integration tests run on both Linux and macOS.
* The setup is greatly simplified (at least from my perspective).
* There was a rebuild before integration tests for some reason — removed it.
* Clippy and build is no longer part of `test` directory (WAT).

A bit of stats regarding build times.

— | Linux | macOS
---|---|---
Before samples removal | 7 min 3 sec | 1 min 13 sec
After samples removal | 3 min 22 sec | 51 sec
After Docker removal | 1 min 20 sec | 2 min 39 sec

macOS started to run integration tests and since Travis most likely uses Mac Mini it took a hit.


artem-zinnatullin:
This is cool and dandy, but I'm not a fan for several reasons:

- How do I run tests locally now without affecting my ssh settings?
- People who want to make small changes would now get rust and other stuff installed on their machines while before they just needed Docker and building Mainframer didn't pollute their local enviornment

ci/build.sh
```diff
@@ -1,80 +1,21 @@
 #!/bin/bash
-set -e
+set -eu
 
-# You can run it from any directory.
-DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+rustup self update
```
artem-zinnatullin, line 4: this installs latest stable version which breaks what I did before — installed specific version of Rust and its tooling

arturdryomov: Travis installs the specific version for you. 

```
curl -sSf https://build.travis-ci.org/files/rustup-init.sh | sh -s -- --default-toolchain=$TRAVIS_RUST_VERSION -y
```

arturdryomov:
@artem-zinnatullin, don’t know how to say this but you are one of not many people who actually use Docker and are familiar enough with it to put everything there and use for development purposes. People would rather install Rust (myself included). Especially if they are familiar enough with Rust. This is a subjective statement of course but I haven’t saw a single person who would use Docker instead of vanilla Gradle, Python, Rust or whatever. Damn, I think I am kind of familiar with Docker and still not a fan of these monstrous scripts around images, especially in the current state.

The issue I agree on is running integration tests locally. I should note that the current setup is weird in any case. Ideally there would be a separate container for the remote machine, not the same one for the local and the remote ones. Unfortunately it is not possible with macOS since there is no way to run Docker here and no way to run CI machines in parallel with link between them.

I can suggest using an exit hook which will mirror constructive actions with destructive ones.

* `ssh-keyscan -t rsa localhost > ~/.ssh/known_hosts` → `ssh-keygen -R hostname`. Or even use `ssh -o StrictHostKeyChecking=no` but you are not a fan of such solutions.
* `cp ~/.ssh/id_rsa.pub ~/.ssh/authorized_keys` → [`grep -v`](https://superuser.com/a/775195)

This should be safe enough taking into account that there are not a lot of people who SSH on the `localhost`.

artem-zinnatullin:
Well I get that Docker is not an easiest thing in the world, but that's how I run tests in Mainframer during its development and this PR breaks this workflow completely.

>Unfortunately it is not possible with macOS since there is no way to run Docker here and no way to run CI machines in parallel with link between them.

This is possible on macOS (there is [official Docker for Mac](https://docs.docker.com/docker-for-mac/) which transparently runs a virtual machine but it's all hidden from you) and I could transform the setup to run 2 separate containers with network connection between them, it just wasn't really necessary yet I agree it's more fair integration test.

It's not safe at all to start an ssh server on developer's machine and mess with its configuration.

I'd rather refactor current setup and optimize build times (btw, 3 minutes is barely an issue given that we don't have that many builds). 

The only reason why current setup is slow on CI is because we don't push the image to any registry so it has to be built every time, locally it actually starts in seconds and skips image build once you have it in cache.

JayNewstrom:
As an outsider, who uses the project, and would contribute if I had an issue or feature request, docker is helpful as it is a tool I'm familiar with, and it would have everything setup for me to run/test the code.

Issue #242, opened 2018-11-28
Strip symbols from release binaries


artem-zinnatullin:
Context: https://twitter.com/artem_zin/status/1067842525317148672

>heh, @\arturdryomov asked if we should rewrite Mainfraimer in Swift, I said it brings giant runtime and indeed as we checked with @\kastiglione, a simple Swift CLI binary is ~10 MB and ~6.5 after strip Mainframer (Rust, release build) is around 725K, 479K after strip

artem-zinnatullin:
Cargo [stabilized `strip` feature](https://doc.rust-lang.org/cargo/reference/profiles.html#strip), we need to look into it and compare results with `strip` util

PR #256, opened 2019-05-01
Add parallel pull mode

### LINKED ISSUE #188: Continuous sync from remote to local machine during remote command execution.

artem-zinnatullin:
Closes #188.

Finally, here is the implementation of new pull mode: `parallel`

In this mode Mainframer pulls files from remote machine in parallel to the remote command execution.

This mode dramatically improves total execution time. 

With `serial` pull mode (1.x, 2.x behavior) user had to wait for all files to be pulled after the remote command finished.

Now with `parallel` pull mode, most of the files will be already pulled by the time remote command finishes and after it finishes last incremental pull will be done (usually ~1 second).

---

Results from `tests/test_pulls_big_files.sh` (creates 30 64MB files):

**1.6x faster overall execution time!**

```console
-------- TEST STARTED (pull mode = 'parallel') test_pulls_big_files.sh --------
:: Mainframer v3.0.0-dev

Pushing...
Push done: took 0 seconds.

Executing command on remote machine (pulling in parallel)...

mkdir build && for ((i=0;i<30;i++)); do dd if=/dev/urandom of=build/buildresult-.txt bs=16M count=4 iflag=fullblock; sleep 1; done;

Execution done: took 41 seconds.
Pulling...
Pull done: took 17 seconds

Success: took 58 seconds.

-------- TEST ENDED (pull mode = 'parallel') test_pulls_big_files.sh --------
```

vs

```console
-------- TEST STARTED (pull mode = 'serial') test_pulls_big_files.sh --------
:: Mainframer v3.0.0-dev

Pushing...
Push done: took 0 seconds.

Executing command on remote machine...

mkdir build && for ((i=0;i<30;i++)); do dd if=/dev/urandom of=build/buildresult-.txt bs=16M count=4 iflag=fullblock; sleep 1; done;

Execution done: took 40 seconds.
Pulling...
Pull done: took 52 seconds

Success: took 1 minute 33 seconds.

-------- TEST ENDED (pull mode = 'serial') test_pulls_big_files.sh --------
```


artem-zinnatullin:
Added results from tests (see PR description)

src/intermediate_config.rs
```diff
@@ -27,7 +29,8 @@ pub struct IntermediatePush {
 
 #[derive(Debug, PartialEq)]
 pub struct IntermediatePull {
-    pub compression: Option<u8>
+    pub compression: Option<u8>,
+    pub mode: Option<PullMode>
```
aneeshusa: nit: use trailing commas

src/intermediate_config.rs
```diff
@@ -92,12 +95,17 @@ fn parse_config_from_str(config_content: &str) -> Result<IntermediateConfig, Str
...
         Yaml::Hash(pull) => {
             let compression = parse_compression(pull, "compression", "pull");
+            let mode = parse_pull_mode(pull);
 
-            match compression {
-                Ok(value) => Some(IntermediatePull {
-                    compression: value,
-                }),
-                Err(error) => return Err(error)
+            if compression.is_err() {
```
aneeshusa: Using the [`?` operator](https://doc.rust-lang.org/book/ch09-02-recoverable-errors-with-result.html) will let you clean this up:  https://play.rust-lang.org/?version=stable&mode=debug&edition=2015&gist=863ccbcfce64499653921e7c339abc11

At some point you may want a more full-fat parsing library, or want to show all parsing errors simultaneously to the user, but for the existing code that playground link should be helpful.

src/sync.rs
```diff
@@ -1,11 +1,44 @@
...
+use crossbeam_channel::Receiver;
+use crossbeam_channel::Sender;
+use crossbeam_channel::unbounded;
+
+use config::Config;
+use ignore::Ignore;
+use remote_command::RemoteCommandResult;
+
+#[derive(Debug, PartialEq, Clone)]
+pub enum PushResult {
```
aneeshusa: Any reason not to make this a [type synonym](https://doc.rust-lang.org/book/ch19-04-advanced-types.html#creating-type-synonyms-with-type-aliases) instead so you can reuse all the functionality of `Result` directly?

aneeshusa: Or actually, how about something like:
```suggestion
pub struct PushResult {
    pub duration: Duration,
    pub result: Result<(), String>,
}
```
I think this would reflect the intent more accurately, looking at `fn push()` below.

aneeshusa (commented):
A few quick comments from a first glance - will try to find time to read through the actual interesting bits e.g. the parallelized bits later :)

src/intermediate_config.rs
artem-zinnatullin: noice, will do!

src/sync.rs
artem-zinnatullin: The only reason to not make a synonym was to better support extension with additional fields, but I think best version would be to combine custom structs with raw Result

```rust
pub struct PushOk {
  pub duration: Duration
}

pub struct PushErr {
  pub duration: Duration
}

Result<PushOk, PushErr>
```

I def feel weird that enums can't have named fields, structs seem to provide more protection from accidental value mismatch

arturdryomov: I guess you meant... `PushEr`...

artem-zinnatullin:
@aneeshusa thanks for review, that was really helpful 👍  

I've addressed all your comments, don't want to waste more of your time but of course you're welcome to review again :)

src/main.rs
```diff
@@ -27,37 +36,76 @@ fn main() {
...
+    match config.pull.mode {
+        PullMode::Serial => println!("Executing command on remote machine...\n"),
+        PullMode::Parallel(_) => println!("Executing command on remote machine (pulling in parallel)...\n")
+    }
+
+    let mut remote_command_readers = remote_command::execute_remote_command(
+        args.command.clone(),
+        config.clone(),
+        sync::project_dir_on_remote_machine(&local_dir_absolute_path.clone()),
+        2
```
aneeshusa, line 69: This `2` is bugging me (perhaps more than it should). Any reason `execute_remote_command` doesn't return the `bus` directly? This will also let you get rid of the `pop`s and their accompanying `unwraps`.

src/main.rs
```diff
@@ -27,37 +36,76 @@ fn main() {
...
+    match pull_result {
+        Err(ref err) => eprintln!("Pull failed: {}, took {}.", err.message, format_duration(err.duration)),
+        Ok(ref ok) => println!("Pull done: took {}", format_duration(ok.duration)),
+    }
 
     match remote_command_result {
-        Err(_) => exit_with_error(&format!("\nFailure: took {}.", format_duration(duration)), 1),
-        _ => println!("\nSuccess: took {}.", format_duration(duration))
+        Err(_) => {
+            match pull_result {
```
aneeshusa: The branches are currently identical, can you get rid of the match?

src/main.rs
```diff
@@ -27,37 +36,76 @@ fn main() {
...
-    if let Err(error) = sync_after_remote_command(&local_dir_absolute_path, &config, &ignore) {
-        exit_with_error(&format!("Sync remote → local machine failed: {}.", error), 1)
+    let remote_command_result = remote_command_readers
+        .pop()
+        .unwrap()
+        .recv()
+        .unwrap();
+
+    match remote_command_result {
+        Err(ref err) => eprintln!("\nExecution failed: took {}.\nPulling...", format_duration(err.duration)),
```
aneeshusa, line 81: The `Pulling...` part of this message seems a bit misleading in the parallel case, as I'd expect one last incremental sync. Consider having separate messaging for the serial vs parallel cases.

src/main.rs
```diff
@@ -27,37 +36,76 @@ fn main() {
...
-        Err(_) => exit_with_error(&format!("\nFailure: took {}.", format_duration(duration)), 1),
-        _ => println!("\nSuccess: took {}.", format_duration(duration))
+        Err(_) => {
+            match pull_result {
+                Err(_)=> exit_with_error(&format!("\nFailure: took {}.", format_duration(total_duration)), 1),
+                Ok(_) => exit_with_error(&format!("\nFailure: took {}.", format_duration(total_duration)), 1),
+            }
+        },
+        Ok(_) => {
+            match pull_result {
```
aneeshusa: Try matching on the tuple `(remote_command_result, pull_result)` to make the whole match more concise.

src/sync.rs
```diff
@@ -28,10 +71,98 @@ pub fn sync_local_to_remote(local_dir_absolute_path: &Path, config: &Config, ign
...
+                        .send(Err(err)) // TODO handle code 24.
+                        .expect("Could not send pull_finished signal");
+                }
+                Ok(_) => thread::sleep(pause_between_pulls),
+            }
+
+            match remote_command_finished_signal.try_recv() {
+                Err(reason) => match reason {
+                    Disconnected => should_run = false,
+                    Empty => thread::sleep(pause_between_pulls)
```
aneeshusa, line 125: Seems like most iterations would actually sleep for 2x `pause_between_pulls`: once above when `_pull` succeeds normally, and once here while the remote command is still executing.

src/sync.rs
```diff
@@ -28,10 +71,98 @@ pub fn sync_local_to_remote(local_dir_absolute_path: &Path, config: &Config, ign
...
+    let (pull_finished_tx, pull_finished_rx): (Sender<Result<PullOk, PullErr>>, Receiver<Result<PullOk, PullErr>>) = unbounded();
+    let start_time = Instant::now();
+
+    thread::spawn(move || {
+        let mut should_run = true;
+
+        while should_run {
+            match _pull(local_dir_absolute_path.as_path(), &config, &ignore) {
+                Err(err) => {
+                    should_run = false;
```
aneeshusa: Do you want to `break` in this case?

Cargo.toml
```diff
@@ -11,3 +11,5 @@ authors = ["Artem Zinnatullin <artem.zinnatullin@gmail.com>", "Artur Dryomov <ar
 [dependencies]
 yaml-rust = "0.4.2"
 linked-hash-map = "0.5.1"
+crossbeam-channel = "0.3"
```
aneeshusa, line 14: I haven't used it in a while, but any reason not to use `std::sync::mpsc`?

src/sync.rs
aneeshusa: IMO it would still be better to have a product type at the top since you'll always want the duration, either via a named struct as I suggested above or an (anonymous) tuple. This also lets you punt on custom structs for the Ok/Err cases until you have extra values there - the compiler will enable safe refactoring later. This lets you avoid adding too much complexity; e.g., the `execute_rsync(&mut command)` on master which is currently in your PR:
```
  match execute_rsync(&mut command) {
        Err(reason) => Err(PushErr {
            duration: start_time.elapsed(),
            message: reason,
        }),
        Ok(_) => Ok(PushOk {
            duration: start_time.elapsed()
        }),
    }
```
could be just
```
  (start_time.elapsed(), execute_rsync(&mut command))
```

Similarly, code like
```
let remote_command_duration = match remote_command_result {
    Err(err) => err.duration,
    Ok(ok) => ok.duration
};
``` goes away.

**Review by aneeshusa — APPROVED** [review]
Overall seems fine, but per https://github.com/buildfoundation/mainframer/pull/256#discussion_r280992815 I see a lot of types being added with an unclear benefit - would love to this feature implemented with fewer lines of code :)

File: src/sync.rs (hunk 3 above)
Comment (artem-zinnatullin) [reply src/sync.rs]: I'd like a product top-level type but I'd also like to immediately be able to distinguish its subtype (if needed)

In Kotlin (best example I have on mind) I'd do this:

```kotlin
sealed class PushResult(val duration: Duration) {
  data class Ok(override val duration: Duration): PushResult(duration)
  data class Err(override val duration: Duration): PushResult(duration)
}
```

This way it's still an enum (sealed class in Kotlin) while also containing always-present data in each case.

In Rust I guess I could do something like this:

```rust
pub enum PushResult {
    Ok(Duration),
    Err(Duration)
}

pub trait _PushResult {
    fn duration(&self) -> &Duration;
}

pub impl _PushResult for PushResult {
    fn duration(&self) -> &Duration {
        match self {
            PushResult::Err(duration) => duration,
            PushResult::Ok(duration) => duration
        }
    }
}
```

wdyt?

src/main.rs
artem-zinnatullin: yeah… I can't return `Bus` (that's what I've tried first until I got compiler yelling at me)

It's owned by closure that I pass to `thread::spawn` in `execute_remote_command`, there is an upstream issue about similar problem https://github.com/jonhoo/bus/issues/19

src/main.rs
artem-zinnatullin: Done, lost some compile-time safety though, but not too bad

src/main.rs
artem-zinnatullin: While it's incremental, it's nontheless *potentially* long

In many cases remote command produces file(s) at the very end, thus parallel pulling might not be parallel

src/sync.rs
artem-zinnatullin: Good catch! I actually only need to sleep while remote command is execution, I think I had that before but then broke it during refactoring, thanks

src/sync.rs
artem-zinnatullin: lemme try to remove `should_run`

Cargo.toml
artem-zinnatullin: Initially I swapped to `crossbeam` because I needed multi-receiver support, but then even that wasn't enough so I've also added `bus`

Decided to keep crossbeam for performance lol (nanoseconds matter hey!), we'll be stripping final binary #242 so it shouldn't matter which one we're using from the binary size point

src/sync.rs
artem-zinnatullin: replaced with `loop`, thanks!

arturdryomov:
@artem-zinnatullin, rebase?

Also expecting a lot of issues with this mode in practice 🤔

artem-zinnatullin:
Rebased, all issues that we will encounter should be solvable

arturdryomov (approved):

clippy.toml
```diff
@@ -0,0 +1 @@
+type-complexity-threshold = 450
```
arturdryomov, line 1: Why do you need it?

artem-zinnatullin:
```console
error: very complex type used. Consider factoring parts into `type` definitions
  --> src/sync.rs:93:47
   |
93 |     let (pull_finished_tx, pull_finished_rx): (Sender<Result<PullOk, PullErr>>, Receiver<Result<PullOk, PullErr>>) = unbounded();
   |                                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: `-D clippy::type-complexity` implied by `-D warnings`
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#type_complexity

error: very complex type used. Consider factoring parts into `type` definitions
   --> src/sync.rs:110:47
    |
110 |     let (pull_finished_tx, pull_finished_rx): (Sender<Result<PullOk, PullErr>>, Receiver<Result<PullOk, PullErr>>) = unbounded();
    |                                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#type_complexity

error: aborting due to 2 previous errors
```

😭

PR #260, opened 2019-05-04
Add content verification to tests

### LINKED ISSUE #257: Verify file content in tests
Both push and pull should guarantee content equality once they're finished.

We can verify sha256 of the files.

artem-zinnatullin:
Closes #257.

Makes sure that actual file(s) content is what we expect after both push and pull in tests.
This is especially important for #256 because it incrementally syncs file(s).


Tagakov (approved):

arturdryomov:
Promised myself to take a look into this after a vacation and of course totally forgot. Sorry mate.

arturdryomov:
@artem-zinnatullin, just a crazy thought — can we replace all this integration testing mess with this?

```
test-something
- local-files
- remote-files-expected
```
This way we can run MF and check the result with the expectation. I strongly feel that we are reinventing a wheel here with an in-house Bash-based (may God have mercy on our souls) testing framework.

artem-zinnatullin:
Yes, very much likely we can, that's a good idea

PR #282, opened 2022-09-26
Swapped custom yaml parsing to object deserialization with serde

ChristopherJMiller:
This PR moves away from yaml parsing to object deserialization. Feature parity was kept with the previous implementation.

Additionally, to use this more modern crate, a project edition had to be set. I chose 2021.

Tests from intermediate_config.rs were kept and reconfigured and now exist in config.rs. It could be argued not all of these need to exist anymore as they can be assumed working by the supporting crates (such as 4 spaces parsing tests)


src/sync.rs
```diff
@@ -25,14 +26,25 @@ pub struct PushErr {
...
-#[derive(Debug, Eq, PartialEq, Clone)]
+#[derive(Debug, Eq, PartialEq, Clone, Deserialize)]
+#[serde(rename_all = "lowercase")]
 pub enum PullMode {
     /// Serial, after remote command execution.
     Serial,
 
     /// Parallel to remote command execution.
     /// First parameter is pause between pulls.
-    Parallel(Duration),
```
ChristopherJMiller, line 35: Duration as a part of `PullMode` was removed due to the config file currently not supporting passing a value for it.  Is this okay to keep or should we figure out a syntax for passing the duration?

Cargo.toml
```diff
@@ -9,10 +9,11 @@ version = "3.0.0-dev"
 authors = ["Artem Zinnatullin <artem.zinnatullin@gmail.com>", "Artur Dryomov <artur.dryomov@gmail.com>", "Mainframer Developers and Contributors"]
 
 [dependencies]
-yaml-rust = "0.4.5"
-linked-hash-map = "0.5.6"
```
akshob, line 13: @ChristopherJMiller I think these should be added back.

ChristopherJMiller: `yaml-rust` is being replaced by `serde`, and I believe `linked-hash-map` doesn't have any usages in the project

src/config.rs
```diff
@@ -1,24 +1,330 @@
...
+    pub fn from_file_contents<'a>(contents: &'a str) -> Result<Self, String> {
+        serde_yaml::from_str::<Config>(contents)
+            .map_err(|err| err.to_string())
+            .and_then(|config| {
+                match (
+                    config.valid_pull_compression_range(),
+                    config.valid_push_compression_range(),
+                ) {
+                    (true, true) => Ok(config),
+                    (false, _) => Err(format!(
```
artem-zinnatullin: While I find this use of matching by two values interesting, I think it really can be replaced with two simple if statements, because this matching won't scale if we need to validate more config values :)

```rust
if ! valid_pull_compression_range {
	err
}

if ! valid_pull_compression_range {
	err
}
```

src/config.rs
```diff
@@ -1,24 +1,330 @@
...
+                    )),
+                    (_, false) => Err(format!(
+                        "'push.compression' must be a positive integer from 1 to 9, but was {}",
+                        config.push.compression
+                    )),
+                }
+            })
+    }
+
+    pub fn valid_pull_compression_range(&self) -> bool {
```
artem-zinnatullin: Do we need these as public functions? Sorry my Rust is getting rusty, in other langs I'd keep these private (except for tests)

src/sync.rs
artem-zinnatullin: Well, we kind of need it, but can def be done in a separate PR — just open an issue if you decide to do it separately :)

artem-zinnatullin (approved):
Looks good! Left few comments

src/sync.rs
ChristopherJMiller: I think we should do it separately to be able to discuss what the config would look like, documented in #284 👍

ChristopherJMiller:
@artem-zinnatullin I think the status checks might be stuck :( would you be able to restart them?

artem-zinnatullin:
Pushed them through, it was a GitHub security measure asking me to approve CI run for first time contributors :)

artem-zinnatullin:
```rust
error: explicit lifetimes given in parameter types where they could be elided (or replaced with `'_` if needed by type declaration)
  --> src/config.rs:25:5
   |
25 |     pub fn from_file_contents<'a>(contents: &'a str) -> Result<Self, String> {
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: `-D clippy::needless-lifetimes` implied by `-D warnings`
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_lifetimes

```

ChristopherJMiller:
Linted :) need another manual push through for the CI check

ChristopherJMiller:
Looks like I don't have access to merge the pull request (needs write access). Everything looks good @artem-zinnatullin if you would like to do the honors :)

artem-zinnatullin:
Merged, thank you for the contribution! I think it does improve the code as you proposed, great job!

